PR #2, opened 2013-11-27
Fix condition on optional Referer header

jeromelebleu:
Fixes access to assets. The Referer header is optional; assets serving will fail if user agent doesn't provide it.


Kloadut:
I'm ok with that, but do you have examples of clients that doesn't provide user-agent ?

jeromelebleu:
IceCat (yes, again!) for example, it has "network.http.sendRefererHeader" set to 1 by default which corresponds to "Send the Referer header when clicking on a link, and set document.referrer for the following page.", so assets are unreachable without this patch...

PR #28, opened 2014-11-27
[fix] Do not limit amount of mail aliases & forward.

opi:


Kloadut:
Is this really working ? :D

(You are such a Lua guru)

opi:
it _is_ working, but we need more testing for sure.

opi:
It may be another issue, but I've run into some cache issue when adding/removing aliases from command line, and try to see what happen on the SSO. I have to restart (not even reload) nginx.

PR #44, opened 2015-07-13
Use PCRE to match configurations patterns

jeromelebleu:
Since Lua regular expressions are quite limited and not common, this PR just makes use of PCRE to match patterns of configurations - such as _redirected_regex_.

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -22,6 +22,9 @@ local conf = config.get_config()
 -- Import helpers
 local hlp = require "helpers"
 
+-- Import Perl regular expressions library
+local rex = require "rex_pcre"
+
 -- Just a note for the client to know that he passed through the SSO
 ngx.header["X-SSO-WAT"] = "You've just been SSOed"
 
@@ -177,9 +180,9 @@ end
 
 if conf["redirected_regex"] then
     for regex, redirect_url in pairs(conf["redirected_regex"]) do
-        if string.match(ngx.var.host..ngx.var.uri..hlp.uri_args_string(), regex)
-        or string.match(ngx.var.scheme.."://"..ngx.var.host..ngx.var.uri..hlp.uri_args_string(), regex)
-        or string.match(ngx.var.uri..hlp.uri_args_string(), regex) then
+        if rex.match(ngx.var.host..ngx.var.uri..hlp.uri_args_string(), regex)
+        or rex.match(ngx.var.scheme.."://"..ngx.var.host..ngx.var.uri..hlp.uri_args_string(), regex)
+        or rex.match(ngx.var.uri..hlp.uri_args_string(), regex) then
             detect_redirection(redirect_url)
         end
     end
@@ -210,8 +213,8 @@ function is_protected()
         end
     end
     for _, regex in ipairs(conf["protected_regex"]) do
-        if string.match(ngx.var.host..ngx.var.uri, regex)
-        or string.match(ngx.var.uri, regex) then
+        if rex.match(ngx.var.host..ngx.var.uri, regex)
+        or rex.match(ngx.var.uri, regex) then
             return true
         end
     end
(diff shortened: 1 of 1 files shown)


Kloadut:
Is this working ?

jeromelebleu:
At the time of this PR, yes! Moreover in some YunoHost apps the use of PCRE is used instead of Lua ones...

opi:
ccan we have both for some times ? in order not to break compatibility ...

jeromelebleu:
@opi no, we can't have both. Anyway, I think - but to be confirmed - that generally packagers are using PCRE naturally...

opi:
@jeromelebleu  I guessed that using `string.match` for Lua regex or `rex.match` for PCRE we can support both regex style, isn't it ?
Also, we need to find every app that use regex to see if backward compatibility is really needed.

jeromelebleu:
Oh, do you mean checking both for PCRE and Lua regex in the `if` conditions? For example:

```
if rex.match(ngx.var.host..ngx.var.uri..hlp.uri_args_string(), regex)
   or rex.match(ngx.var.scheme.."://"..ngx.var.host..ngx.var.uri..hlp.uri_args_string(), regex)
   or rex.match(ngx.var.uri..hlp.uri_args_string(), regex)
   or string.match(ngx.var.host..ngx.var.uri..hlp.uri_args_string(), regex)
   or string.match(ngx.var.scheme.."://"..ngx.var.host..ngx.var.uri..hlp.uri_args_string(), regex)
   or string.match(ngx.var.uri..hlp.uri_args_string(), regex) then
```

Why not... Anyway, I think we could improve those conditions to reduce it _cost_.

opi:
> Oh, do you mean checking both for PCRE and Lua regex in the if conditions?

exactly ! not too ugly ? we might document that PCRE is the way to go, and Lua regex are deprecated, without breaking anything.

zamentur:
I was not aware of this PR. It's undocumented.

zamentur:
Expressions currently use by apps (listed in community.json, YunoHost-Apps and labriqueinternet):

- "$domainregex/jsonrpc.*$,$domainregex/image.*$" (kodi)
- "$(sed 's/[\.\-]/\%&/g' <<< $domain)/%.well%-known/.*" (nextcloud)
- "$domain/%.well%-known/acme%-challenge/.*$" (letsencrypt_ynh)
- "$domain_regex$path/login$","$domain_regex$path/logout$","$domain_regex$path/api$","$domain_regex$path/extensions$","$domain_regex$path/stats$","$domain_regex$path/d/.*$","$domain_regex$path/a$","$domain_regex$path/$" (lstu)
- https://github.com/YunoHost-Apps/bozon_ynh/blob/master/scripts/install#L90 (bozon)
- "$domain_regex$path_url/stats$","$domain_regex$path_url/manifest.webapp$","$domain_regex$path_url/$","$domain_regex$path_url/d/.*$","$domain_regex$path_url/m/.*$" (lutim)
- "$domain_regex$path_url/$","$domain_regex$path_url/admin.php.*$" (jirafeau) 
- "$domain_regex$path/stats$","$domain_regex$path/manifest.webapp$","$domain_regex$path/$","$domain_regex$path/d/.*$","$domain_regex$path/m/.*$" (lufi)

Avec domain_regex=$(echo "$domain" | sed 's@-@.@g')

Nextcloud, bozon and letsencrypt works only with lua pattern:
- "$(sed 's/[\.\-]/\%&/g' <<< $domain)/%.well%-known/.*" (nextcloud)
- "$domain/%.well%-known/acme%-challenge/.*$" (letsencrypt_ynh)
- https://github.com/YunoHost-Apps/bozon_ynh/blob/master/scripts/install#L90 (bozon)

zamentur:
I had updated this code, i close this pr as the original branch doesn't exists anymore

https://github.com/YunoHost/SSOwat/pull/102

PR #55, opened 2016-04-21
Regex non reconnues sur protected_regex

maniackcrudelis:
Les patterns sont interprétés correctement sur unprotected_regex, mais pas sur protected_regex.
L'ajout de ..hlp.uri_args_string() corrige ça et permet d'interpréter correctement les patterns

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -210,8 +210,8 @@ function is_protected()
         end
     end
     for _, regex in ipairs(conf["protected_regex"]) do
-        if string.match(ngx.var.host..ngx.var.uri, regex)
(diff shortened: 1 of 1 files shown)


opi:
seems good to me; as `hlp.uri_args_string()` is present everywhere else (redirected, skipped, unprotected), consider adding this to `protected_urls` too.

opi:
and then, could you provide an example of app/regex, I'd like to test but don't really know how

opi:
Merged in 3b2bc73df5d8f62c67f885097ffc3fe80029e257

PR #57, opened 2016-04-30
[fix] Skipped url should be really ignored.

opi:


alexAubin:
Do you know what's the status of this ? It might fix https://dev.yunohost.org/issues/717

access.lua
```diff
@@ -227,25 +227,30 @@ end
...
-        or  hlp.string.starts(ngx.var.uri..hlp.uri_args_string(), url))
-        and not is_protected() then
-            return hlp.pass()
+function is_skipped()
+    if conf["skipped_urls"] then
+        for _, url in ipairs(conf["skipped_urls"]) do
+            if (hlp.string.starts(ngx.var.host..ngx.var.uri..hlp.uri_args_string(), url)
+            or  hlp.string.starts(ngx.var.uri..hlp.uri_args_string(), url))
+            and not is_protected() then
+                hlp.pass()
```
opi, line 236: this should not be a part of this function. `is_skipped` must only return true or false

access.lua
```diff
@@ -227,25 +227,30 @@ end
...
-        if (string.match(ngx.var.host..ngx.var.uri..hlp.uri_args_string(), regex)
-        or  string.match(ngx.var.uri..hlp.uri_args_string(), regex))
-        and not is_protected() then
-            return hlp.pass()
+    if conf["skipped_regex"] then
+        for _, regex in ipairs(conf["skipped_regex"]) do
+            if (string.match(ngx.var.host..ngx.var.uri..hlp.uri_args_string(), regex)
+            or  string.match(ngx.var.uri..hlp.uri_args_string(), regex))
+            and not is_protected() then
+                hlp.pass()
```
opi, line 247: same as before

access.lua
```diff
@@ -344,14 +349,15 @@ if auth_header then
...
 --
 
-hlp.flash("info", hlp.t("please_login"))
-local back_url = ngx.var.scheme .. "://" .. ngx.var.host .. ngx.var.uri .. hlp.uri_args_string()
-return hlp.redirect(conf.portal_url.."?r="..ngx.encode_base64(back_url))
+if not is_skipped() then
+    hlp.flash("info", hlp.t("please_login"))
+    local back_url = ngx.var.scheme .. "://" .. ngx.var.host .. ngx.var.uri .. hlp.uri_args_string()
+    return hlp.redirect(conf.portal_url.."?r="..ngx.encode_base64(back_url))
+end
```
opi, line 363: `else hlp.pass()`

alexAubin:
[Semi-automatic message] 

PR #58, opened 2016-05-02
Move app details in conf to a separate table

gkrnours:
Apps details in conf are moved from the users table to a separate apps table.
More details are available in the yunohost pull request [https://github.com/YunoHost/yunohost/pull/144](144) and yunohost issue [https://dev.yunohost.org/issues/302](302)

PR #59, opened 2016-05-11
provide more details about apps to template

gkrnours:
#58 got deleted by accident while trying to make it compatible with current unstable branch. So same content and ready to merge.

### DIFF
--- a/conf.json.example
+++ b/conf.json.example
@@ -17,14 +17,26 @@
        "Email": "mail",
        "Name": "cn"
    },
-   "users": {
-       "myuser": {
-          "example.com/myapp": "My App",
-          "example.com/myapp2": "My second App"
+   "apps": {
+       "app_id": {
+           "id": "app_id",
+           "name": "App Name",
+           "url": "example.com/app",
+           "description": "The app description"
        },
-       "myuser2": {
-          "example.org/myapp": "My other domain App",
-          "example.com/myapp2": "My second App"
+       "second_app": {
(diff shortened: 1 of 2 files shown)


julienmalik:
Thanks for the contribution. Looks good, but I want to test it. Anyway this should be reviewed by at least @opi

julienmalik:
Didn't notice https://github.com/YunoHost/yunohost/pull/144 and https://dev.yunohost.org/issues/302 at the time I answered, sorry

gkrnours:
A bit more than test might be required as the patch alter the format expected for conf.json without providing any fall back

alexAubin:
I'm just randomly reading this PR and yunohost#144. In an attempt to revive this, is there anything needed to implement, apart from reviews/tests ? We should definitely be cautious with this, but it kinda looks like it's a relatively small, safe and legit change.

alexAubin:
Random bump in case someone has a clue about this :s

Psycojoker:
The idea was to modify sso template after that to display more information in it iirc, there was a screenshot somewhere but it looks like lost it because redmine sucks x(

gkrnours:
Before merging, there is a decision to take on if we want to change the configuration format. 

That might not be the best choice.

Maybe it would be better to have a configuration being a list of app and a second file being the apps definition with all the details. That might provide a smoother experience when shipping the code and simplify internationalisation.

alexAubin:
[Semi-automatic message] 

PR #60, opened 2016-05-18
CSS changes to follow transition effects guidelines 

Acombes:
- Reordered the properties to be able to look through the rules more easily
- Changed the tiles animation from using `top`/`left` properies to using performance optimized `transform` property.
- Replaced all the `transition: all;` properties with more narrow, better suited transition-property values


opi:
oh, so much diff, it will be hard to review.. but thanks for impulse some cleaning of these CSS.

> Reordered the properties to be able to look through the rules more easily

what is your guidelines here ?

Acombes:
> what is your guidelines here ?

The gist of it is to group the properties by groups (e.g. properties affecting the box model, properties affecting the box positioning, text properties, etc.) and to keep them in the same order. That allows someone used to this method to quickly go through the properties when searching for a specific one.

The concept of this is a generic one, but then, one could argue that the implementation i'm using is more of a personal one, but I started using it before actually hearing about such methods. I'm usually going this way:
1. Box model: `display`, `width`, `height`
2. Box positioning: `position`, `top`, `right`, `bottom`, `left`, `z-index`
3. Box styling: `border`, `background`
4. Text formatting: `font-style`, `font-weight`, `font-size`, `font-family`, `text-align`, ...
5. Box transformations: `transform`
6. Transitions / Animations

That's not an exhaustive list, but the general idea is to order the properties from the more fundamental ones (the ones that will strongly affect how the element is displayed) to the lighter ones (how the text inside is displayed) with an exception for transformations and transitions because they usually come with vendor prefixes, and I find it prettier not to have the chunks of identical properties in the middle of the block.

As a side note, there are some solutions that comb through the stylesheets to reorder them:
https://github.com/csscomb/csscomb.js/tree/master/config

alexAubin:
[Semi-automatic message] 

PR #67, opened 2016-11-15
[fix] Refresh ldap info before loading page that requires it, fixes #633

julienmalik:
Corresponding ticket [dev@#633](https://dev.yunohost.org/issues/633).

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -274,12 +274,25 @@ function authenticate(user, password)
     end
 end
 
+function delete_user_info_cache(user)
+    cache:delete(user.."-"..conf["ldap_identifier"])
+    local i = 2
+    while cache:get(user.."-mail|"..i) do
+        cache:delete(user.."-mail|"..i)
+        i = i + 1
+    end
+    local i = 2
+    while cache:get(user.."-maildrop|"..i) do
+        cache:delete(user.."-maildrop|"..i)
+        i = i + 1
+    end
+end
 
 -- Set the authentication headers in order to pass credentials to the
 -- application underneath.
 function set_headers(user)
 
-    -- We definetly don't want to pass credential on a non-encrypted
+    -- We definitely don't want to pass credentials on a non-encrypted
     -- connection.
     if ngx.var.scheme ~= "https" then
         return redirect("https://"..ngx.var.host..ngx.var.uri..uri_args_string())
@@ -287,7 +300,7 @@ function set_headers(user)
 
     local user = user or authUser
 
-    -- If the password is not in cache of if the cache has expired, ask for
+    -- If the password is not in cache or if the cache has expired, ask for
     -- logging.
     if not cache:get(user.."-password") then
         flash("info", t("please_login"))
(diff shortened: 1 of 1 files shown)


helpers.lua
```diff
@@ -471,6 +484,8 @@ function get_data_for(view)
         or view == "edit.html"
         or view == "password.html"
         or view == "ynhpanel.json" then
```
opi, line 486: `ynhpanel.json` is called on every page for every app, so having some cache there seems a good idea, isn't it ?

julienmalik: do you know why it contains the aliases & maildrops ?

opi: I don't know. this is useless. Required informations are uid, givenName, sn, cn and mail.
maildrop and mailaliases are not needed.

julienmalik: OK then let's fix that in this PR too.

mbugeia: Maildrop is used by email transfert feature, I don't know if it's relevant here.

julienmalik: Hum, second thought : let's do what you proposed first : take ynhpanel.json out of the "needs cache refresh" list. Since it does not need aliases and forwards, it is not that much important if these info are out of date.

opi: agreed

Psycojoker:
Juste pour dire que j'ai pas d'avis sur le code de cette PR (je connais pas bien ssowat), je vous fais confiance pour merger :)

Ah et sur le but je suis tout à fait d'accord.

helpers.lua
```diff
@@ -485,7 +485,15 @@ function get_data_for(view)
...
 
-        delete_user_info_cache(user)
+        -- Invalidate cache before loading these views.
+        -- Needed if the LDAP db is changed outside ssowat (from the cli for example).
+        -- Not doing it for ynhpanel.json only for performance reasons,
+        --   so the panel could might wrong first name, last name or main email address
+        if view ~= "ynhpanel.json" then
```
opi: I'd rather prefer to list the routes where we wan't to invalidate cache, for readability, but it's really a personnal point of view.

helpers.lua
```diff
@@ -485,7 +485,15 @@ function get_data_for(view)
         or view == "password.html"
         or view == "ynhpanel.json" then
 
-        delete_user_info_cache(user)
+        -- Invalidate cache before loading these views.
+        -- Needed if the LDAP db is changed outside ssowat (from the cli for example).
+        -- Not doing it for ynhpanel.json only for performance reasons,
+        --   so the panel could might wrong first name, last name or main email address
```
opi: This wording need a fix ;)

julienmalik: Fixed the wording

helpers.lua
julienmalik: I did not change the code, since just before we have an enumeration of all routes, so IMHO it respects the surrounding code convention.
Also I don't know enough lua yet, so I don't know the simple way to do what you propose :)

PR #70, opened 2016-12-10
Agentcobra skin patch

agentcobra:
add skin support


agentcobra:
It's running on my ynh (https://agentcobra.net)

zamentur:
Thanks for your contribution !
The idea is pretty cool !
Your pull request will be reviewed by several developers before to be intregrated.

I think the "brique" image license isn't compatible to be packaged with ssowat (AGPLv3). So we should change it.

About the code, some files (like translation) seems unchanged, it could be convenient to find a way to avoid to duplicate this kind of files.

Psycojoker:
Small css bug, you need to add:

```css
html {
    width: 100%;
    height: 100%;
}
```

To your css file for the "background: cover" to streach itself on big screens.

opi:
My point of view on this PR is that the "brique theme" must be separated from the "make ssowat themeable" code. The internetcube project should take care of its theme apart from SSOwat.  

I think we should close this PR, and work on a proper way to handle SSOwat skin (or theme).

- [ ] Move locales to avoid duplicates
- [ ] Handle mustache files https://github.com/YunoHost/SSOwat/blob/unstable/helpers.lua#L425-L437
- [ ] Decide if we wan't full theme override, or only CSS (+img) skin
- [ ] Allow theme (or skin) to be in any absolute path on the server ?

( don't think that I don't like your PR, I thank you to bootstrap this long-wanted feature ! )

julienmalik:
Would shipping ssowat theme/skin as standard yunohost apps make sense ?

agentcobra:
>  - [ ]  Allow theme (or skin) to be in any absolute path on the server ?
- [ ] Add a gallery with easy way to download skin ?

Psycojoker:
Introducing theme support would be indeed very interesting, but that's another level of work.

a1ex4:
You also need to take into account the [path to locales here in init.lua](https://github.com/YunoHost/SSOwat/blob/unstable/init.lua#L39).

agentcobra:
I can (for the moment) use the locales from portal folder and later reorganize the ssowat folder's structure

alexAubin:
[Semi-automatic message] 

agentcobra:
@alexAubin ok, no problem

PR #71, opened 2016-12-21
[fix] Fix tile not displayed when app is installed on root (bug #285)

JimboJoe:
…p is installed on root path)

The issue is systematic when an app is installed on root path and that path is set in unprotected_urls. It is caused by the unprotected_urls processing in access.lua that passes the URL before having the chance to process specific files (including YunoHost tile).
The proposed fix changes the processing order in access.lua to process specific files (including YunoHost tile) before unprotected_urls.

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -187,7 +187,40 @@ end
 
 
 --
--- 4. Protected URLs
+-- 4. Specific files (used in YunoHost)
+--
+-- We want to serve specific portal assets right at the root of the domain.
+--
+-- For example: `https://mydomain.org/ynhpanel.js` will serve the
+-- `/yunohost/sso/assets/js/ynhpanel.js` file.
+--
+
+if hlp.is_logged_in() then
+    if string.match(ngx.var.uri, "^/ynhpanel.js$") then
+        hlp.serve("/yunohost/sso/assets/js/ynhpanel.js")
+    end
+    if string.match(ngx.var.uri, "^/ynhpanel.css$") then
+        hlp.serve("/yunohost/sso/assets/css/ynhpanel.css")
+    end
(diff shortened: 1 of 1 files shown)


opi:
oh thanks, I'll definitely try to test it asap !

JimboJoe:
I hadn't tested the administration panel... and it wasn't working. I adapted the fix to handle specific files  before unprotected_urls and AFTER skipped_urls. So far, so good!

M5oul:
I tested it and I wasn't able to display YunoHost tile on [duniter](https://github.com/duniter/duniter_ynh) and jirafeau apps which were installed on root path.

JimboJoe:
I did some more tests: 
* it doesn't fix the problem for: Nextcloud, Jirafeau, NetData
* it fixes the problem for: wallabag2, dokuwiki, Grafana

I think it has to deal with the different Nginx configurations. From my understanding, the nginx configurations are applied before `access.lua` in applied.

alexAubin:
So does this mean thas 1) this PR is still good for merging and apps like nextcloud, jirafeau, duniter should be fixed ? Or 2) does this PR requires some work to make it work also for these ?

(To me it looks like it's the first)

JimboJoe:
As mentioned in the [original bug](https://dev.yunohost.org/issues/285), this PR is still relevant and I confirm it corresponds to your first proposition :-)
I can also add it's only a partial fix for the bug.

alexAubin (approved):
I have not tested the PR, but following the discussion on the [original bug](https://dev.yunohost.org/issues/285) and other PRs, it LGTM

zamentur (approved):
reviewed and tested with wallabag2

alexAubin:
Annnnnd sounds like we're ready to merge /o/

PR #73, opened 2017-01-16
Fix HTTP cookie caching

JimboJoe:
- Use "Expires" instead of "Max-Age" when using a cookie date (Max-Age is used with an interval of seconds in the future: https://en.wikipedia.org/wiki/HTTP_cookie#Expires_and_Max-Age)
- Fix cookie dates to be compliant with specifications

Fixes errors with various "picky" clients (for example, Lightroom/Piwigo plugin).

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -131,11 +131,11 @@ end
 function delete_cookie()
     conf = config.get_config()
 
-    expired_time = "Thu, Jan 01 1970 00:00:00 UTC;"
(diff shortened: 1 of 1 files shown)


opi:
looks good, I'd like to test it anyway

opi:
Merged with fb99ee217793f6442576c91f88bfd69f49baf660

opi:
Note that I decided to use 'Expires' too in cookie creation : a2af42144bd84a83ed23309be1094aab6f846b25

PR #80, opened 2017-05-15
uses hmac_sha512 for hasing the token and don't store the key in it anymore

Psycojoker:
This remove the old really really bad algo, to say the least, for hashing cookies and uses hmac_sha512 instead. The official recommendation for that says we should uses sha256 but sha512 was 3 chars ago and I didn't why I shouldn't uses it.

I'm really not proud of the current implementation but this is a result of the total disaster that is the current lua ecosystem as explained in the comment.

I finally chose a solution that was working out of the box from code that I can trust (randomly found on the internet lua library wasn't really trustworthy...) and didn't require any new dependencies. The result still sucks but I didn't wanted to became mad and anything moving away from the current situation seemed good.

I've tried to uses stdin/stdout of the program to avoid giving him the information has arguments for calling, but lua being the disaster that it is, isn't able to do bidirectional communication with another process. We could uses a temporary file but it's only delete at the end of the process...

I'm asking a friend for review. It should be good to have the opinion of someone else from security™ (like nicoo).


athoune:
We are all gonna die.

athoune:
Hum, something like :
```
apt-get install luarocks libssl-dev
luarocks install luacrypto
```

With a pinned version, not globally installed. It's the dependency of resty-hmac, and it's as secure as OpenSSL.

opi:
yet another dependency manager ...

Psycojoker:
Discussing with could also uses openssl directly to remove python script which would already be a step forward https://stackoverflow.com/questions/7285059/hmac-sha1-in-bash That would also mean adding openssl as a dependency if it's not already the cases.

The solution to uses luacrypto seems somehow "the right way" to do it, but packaging and shipping everything for debian is something that would take me days of work and I'm not convinced I would do a good job at it so I'm not really motivated by this direction.

athoune:
@opi Debian package as universal dependency manager is a joke, system package are for system libs. There is a dh-lua, so, a real debian package is doable.

opi:
@athoune It was a bad monday morning joke ;)  I know Debian can't solve every issue ;)

More seriously, I don't have anything against using luarocks to get the luacrypto tool. But we have to find a way to handle this dependency in our building process.

opi:
By looking at https://github.com/mkottman/luacrypto, we can forget the easy way, aka "Copy/paste a magic `crypto.lua` file locally" :/

Psycojoker:
Yes, I've been looking at a lot of crypt lua lib. Lua lockbox is the only ones that do everything in full lua but ... it doesn't work \o/ (for hamc_sha* at least, the other parts seems fine).

Also using randomly found on the internet code for crypto really doesn't talk to me :/

Psycojoker:
As discovered by opi, lua-crypto only supports up to sha2 so we can't uses it.

Psycojoker:
I've switched to using openssl directly.

alexAubin (approved):
Don't know much about the security of SSOwat so I have no strong opinion on this, but this looks legit to me.

Psycojoker:
I've discussed this PR with a friend "in security" ™ and:

* while this PR is "cracra" it is acceptable and can be merged
* the actual code isn't that disastrous
* he found https://code.google.com/archive/p/sha2/ which produce the same hmac_sha512 than openssl
* the "leak those information in the process list" is acceptable because if someone can read that it's already too late for this

So, we can ever:

* directly merge this PR
* see how we can integrate "sha2" into YunoHost

Sha2:
* works on ynh-dev AND on arm
* NEEDS to be installed with luarocks which we need to install then run in the debian package
* we can see if we can make our how debian package for that

Personally I would merge this PR because: 
* I'm lazy and don't want to suffer doing work on lua again (it's really not fun at all)
* I don't want to wait too much and would be happy to release those security improvement soon combined with the previous security release

alexAubin:
I'm okay with merging in the current state (though, again, don't know much about what's the risk exactly).

Regarding the implementation of sha2, is it an option to include this as a "vendor" thing, just like acme-tiny ?

Psycojoker:
> Regarding the implementation of sha2, is it an option to include this as a "vendor" thing, just like acme-tiny ?

No, it needs compilation :(

opi (approved):
don't know if it's better, I guess so ! But at least my ssowat still works after using this patch !

opi:
as reported by some users, the command `echo -n 'message' | openssl sha512 -hmac 'key' ` works well on various devices/architecture (baremetal, vagrant box, Olinuxino, RPi, BananaPi, ...)

Other pull-requests with security fixes and improvments are going to be merged today, so I'll merge this PR as-is, even if it's not perfect.

PR #81, opened 2017-05-17
Remove ip address from auth token, it's not needed and this fix infinite redirection

Psycojoker:
Hello,

THIS PR IS BASED ON #80 since it changes the same part of the code. YOU SHOULD ONLY LOOK AT THIS COMMIT https://github.com/YunoHost/SSOwat/pull/81/commits/0c079c8502896258a91c32769d3a40017cfa1fff

After a discussion with a friend in the security field regarding #79 and #80 he told me that putting the IP in the token is useless here because:

* it protects from stealing the token
* but the token is always on https since it's marked as secured
* this mean that if someone wants to steal it it need to access the user computer
* this is a situation in which the user is in a really bad situation and have way other problems than yunohost token

This will also fix the "infinite redirection" problem that a signifiant number of users experienced.

Code has been tested and works for me.


alexAubin (approved):
Looks good to me

opi (approved):

opi:
manually merged with 98a6879ab4d39e1130ec29393c04e50663735c81

PR #82, opened 2017-05-18
[fix] CDA security issue

opi:
https://dev.yunohost.org/issues/921


opi:
Approved by @Psycojoker during XMPP review

PR #84, opened 2017-05-22
[enh] uses caching for hash to avoid heavy recalculation and process spawning

Psycojoker:
@sidddy can you tell me if this PR improve perfs on your side? It should be way faster now.

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -72,23 +72,31 @@ end
 
 -- Hash a string using hmac_sha512, return a hexa string
 function hmac_sha512(key, message)
-    -- lua ecosystem is a disaster and it was not possible to find a good
-    -- easily multiplatform integrable code for this
-    -- Python has this buildin, so we call it directly
-    --
-    -- this is a bad and probably leak the key and the message in the process list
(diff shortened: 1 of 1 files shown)


opi (commented):
seems fair. need to test though

sidddy:
I did a little performance test for loading a random application (shellinabox) from my raspi (from a remote location, so the response times will have some fluctuation, but anyway the conclusion is obvious... ;-) )

- via Basic Auth (no SSOWat involved): 741ms
- via SSOWat using old MD5 hash: 765ms
- via SSOWat using SHA512: 1006ms
- via SSOWat using cached SHA512: 747ms

Especially for shellinabox, roundtriptimes are crucial. In my setup, the RTT is around 50ms for basic auth, md5 and cached sha512. Using un-cached sha512, the RTT is around 130ms, which gives a really bad experience for shellinabox.

So, conclusion: I like the new sha512 cache :-)

opi (approved):
thanks

Psycojoker:
Thanks a lot for your testing sidddy <3

PR #86, opened 2017-08-10
Fixed support for incomplete translations 

MCMic:
Fallback to default language for missing strings.

Otherwise buttons and links are empty in case of incomplete translation (had the problem with esperanto and hindi).
You may want to test if it works when conf.lang is not defined but I think it should work as well.

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -30,7 +30,7 @@ end
 
 -- Get the index of a value in a table
 function index_of(t,val)
-    for k,v in ipairs(t) do 
+    for k,v in ipairs(t) do
         if v == val then return k end
     end
 end
@@ -50,8 +50,8 @@ end
 
 -- Find a string by its translate key in the right language
 function t(key)
-   if conf.lang and i18n[conf.lang] then
-       return i18n[conf.lang][key] or ""
+   if conf.lang and i18n[conf.lang] and i18n[conf.lang][key] then
+       return i18n[conf.lang][key]
    else
        return i18n[conf["default_language"]][key] or ""
    end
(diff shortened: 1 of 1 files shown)


Psycojoker:
Need to be tested but code looks good from here, thx for this PR :)

opi (commented):
Looks good. Untested because it's a smartphone review ;)

alexAubin:
Tested and works as expected :+1: . This is actually kind of a serious UI/UX bug.

Can't really tell much about the code though because I don't know much of lua/ssowat ;)

But if people are happy with the code I guess we can merge

Psycojoker:
Merging then since code is good for me and opi.

Psycojoker:
It appears that this PR is provoking this exception on our CI when testing on a private installation of several apps for an unknown reason.

```
2017/08/18 17:34:59 [error] 1165#0: *2 lua entry thread aborted: runtime error: /usr/share/ssowat/helpers.lua:560: attempt to index a nil value
stack traceback:
coroutine 0:
        /usr/share/ssowat/helpers.lua: in function 'get_data_for'
        /usr/share/ssowat/helpers.lua:470: in function </usr/share/ssowat/helpers.lua:419>, client: 10.1.4.1, server: domain.tld, request: "GET /yunohost/sso/?r=aHR0cDovL3NvdXMuZG9tYWluLnRsZC8= HTTP/1.1", host: "domain.tld"
```

We are investigating but this is really weird. This seems to be the location of the bug https://github.com/YunoHost/SSOwat/pull/86/files#diff-843bc3c9d89e152601a2d4b75748fc56R560 but I'm not able to reproduce this on my testing env.

Psycojoker:
It's fixed.

PR #87, opened 2017-08-15
[fix] Auto-update user password hashes with new algo

Psycojoker:
Related to hash PR on yunohost

Related to https://github.com/YunoHost/yunohost/pull/350

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -81,7 +81,7 @@ function hmac_sha512(key, message)
         -- this is really dirty and probably leak the key and the message in the process list
         -- but if someone got there I guess we really have other problems so this is acceptable
         -- and also this is way better than the previous situation
-        local pipe = io.popen("echo -n '" ..message.. "' | openssl sha512 -hmac '" ..key.. "'")
+        local pipe = io.popen("echo -n '" ..message:gsub("'", "'\\''").. "' | openssl sha512 -hmac '" ..key:gsub("'", "'\\''").. "'")
 
         -- openssl returns something like this:
         -- root@yunohost:~# echo -n "qsd" | openssl sha512 -hmac "key"
--- a/helpers.lua
+++ b/helpers.lua
@@ -293,6 +293,7 @@ function authenticate(user, password)
     -- cache shared table in order to eventually reuse it later when updating
     -- profile information or just passing credentials to an application.
     if connected then
+        ensure_user_password_uses_strong_hash(connected, user, password)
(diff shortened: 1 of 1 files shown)


helpers.lua
```diff
@@ -808,6 +837,16 @@ function edit_user()
     end
 end
 
+-- hash the user password using sha-512 and using {CRYPT} to uses linux auth system
+-- because ldap doesn't support anything stronger than sha1
+function hash_password(password)
+    -- TODO is the password checked by regex? we don't want to
+    -- allow shell injection
+    local mkpasswd  = io.popen("mkpasswd --method=sha-512 '" ..password.."'")
```
alexAubin: Does this shows up in `ps -ef` ? Do we care about it ? (Maybe this is the only acceptable solution since we don't have a propre lua lib for that :/)

Psycojoker: Yes it does ... and yes, we don't have a choice right now :(
It happens very unfrequently and tbh, we already store in permanence the user passwords un nginx memory cache to reconnect to ldap if we lose connection -_-' (I didn't managed to find an acceptable way to fix that)

alexAubin: [Okay](http://yolocowboy.acteurdu.net/okay.ogg) :s

Psycojoker: Well, technically there is this thing that exists but it hasn't been updated since ... 2004 and it's not packaged in debian (so it will probably never be) http://files.luaforge.net/releases/lua-crypt/lua-crypt/lua-crypt1.0

We could experiment and migrate to that later on once we managed to take back control on our debian toolchain, but that's not for today and it's more urgent to fix this.

alexAubin (commented):
Tested, works as expected. Made a quick review (not a security expert). I tested a few injection through the password (c.f. the `io.open()`). Haven't been able to inject anything but we should have a deeper look before merging though...

Psycojoker:
I have made tests regarding the shell injection and actually it seems that we are safe because we quote variables using '' when using `io.popen`. I've tested both <code>``</code> and `$()` syntaxes.

```lua
> a = "`echo qsd > pouet`"
> io.popen("echo '" .. a .. "'")  -- not vulnerable
> io.popen("echo " .. a) -- vulnerable
```

But we were vulnerable to injection using `'` so I've added a regex for that according to https://github.com/ncopa/lua-shell/blob/c1e1c0eebdf220bf934a183e1fd030f83dd80866/shell.lua#L9 (and made some tests of course and also retested the login)

PR #92, opened 2017-10-03
1º Let the admin choose if the passwords should be overwriten using crypt; 2º minor fixes

tYYGH:
1º Let the admin choose if the passwords should be overwriten using crypt:

* new _optional_ conf parameter `ldap_enforce_crypt` with _default value_ `true` => *no regression* on existing setups
* only rehash the password using crypt if the above parameter is true.

2º Minor fixes

* In function `ensure_user_password_uses_strong_hash`, the LDAP parameters are used instead of being hard-coded.
* Set a default value to `skipped_urls` (=`{}`) so that the README can come back to its original state, ie. only `portal_domain` is mandatory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ If you use YunoHost, you may want to edit the `/etc/ssowat/conf.json.persistent`
 
 ## Available parameters
 
-These are the SSOwat's configuration parameters. Only `portal_domain` and `skipped_urls` are required, but it is recommended to know the others to fully understand what you can do with SSOwat.
+These are the SSOwat's configuration parameters. Only `portal_domain` is required, but it is recommended to know the others to fully understand what you can do with SSOwat.
 
 #### portal_domain
 
@@ -103,6 +103,10 @@ LDAP user identifier (**default**: `uid`)
 #### ldap_attributes
 
 User's attributes to fetch from LDAP (**default**: `["uid", "givenname", "sn", "cn", "homedirectory", "mail", "maildrop"]`)
+
+#### ldap_enforce_crypt
+
+Let SSOwat re-encrypt weakly-encrypted LDAP passwords into the safer sha-512 (crypt) (**default**: `true`)
 
 #### allow_mail_authentication
 
--- a/README.md
+++ b/README.md
@@ -74,11 +74,11 @@ Domain of the authentication portal. It has to be a domain, IP addresses will no
 
 #### portal_path
 
-URI of the authentication portal (**default**: `/ssowat`)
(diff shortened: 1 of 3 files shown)


tYYGH:
Hum… the PR mechanism does not seem to be based on diff / patches, but on branches / merging instead. Result: the previous PR is not taken into account :-/
Not a big deal this time, but I fear I’ll have to rewrite my history for the next PRs…

alexAubin:
Well, about the "enforce bcrypt" thing, I wonder what should be done.

These lines are tightly coupled to YunoHost and were introduced more as a one-time mechanism to transparently migrate password for users, rather than a mid/long-term need. Imho this is bad practice (though it was kinda necessary in our case) to rewrite the password hash in the LDAP db... because it's not SSOwat's job to manage how password are stored in db. It's the user management system (which in our case is basically the `yunohost user` command).

So imho, it should be more something like : 

```python
# This is legacy code - should be removed on the long term.
if yunohost:
    ensure_user_password_uses_strong_hash
```

(not actual implementation proposal, just a sketch)

tYYGH:
@alexAubin I felt the same way about this rewriting not being Ynh’s job, but the admin’s instead. That is why I posted this PR. As I see things, if current Ynh installations indeed do not need this:

* either this legacy code is not needed anymore, and should just be removed,
* or it is a transient setting, and then we could change the default value to `false` instead of `true`.

What do you think?

[Edit:] By the way, is your “`if yunohost`” pseudo-code based on a real variable? I mean: is there really an easy and fast way to know wether or not we are in YunoHost?

alexAubin:
Well it is still needed because this is a pretty recent change (dates back from just one month, so this is probably still relevant for a lot of instances...). And we don't have a schedule to remove this piece of code anytime soon ...

But the way I see things, if SSOwat really is to be developed as an independent block (which I agree with, because it's interesting both for people willing to use this outside YunoHost, and for YunoHost since this brings more people to contribute / review / test the code and make it more generic etc, ...), I think we (the YunoHost people) should have our own "patched" version with this piece of code which is relevant for us but not for other people...

tYYGH:
@alexAubin If it’s OK with you and other YunoHost contributors, I don’t mind leaving things as I put them in my branch. Indeed, although _I_ am not interested in this pwd rewriting, I would understand if some admins were; it is after all a step towards a more secure setup…
(and the place taken in the source code is correct)

alexAubin (approved):
Trying to decide what to do with this ... it looks safe and relevant to merge this.

Proposing to merge in a few days ?

alexAubin:
Thank you for your contribution and sorry for the delay :wink:

PR #95, opened 2017-11-18
[fix] Add whois as dependency

alexAubin:
c.f. https://forum.yunohost.org/t/nging-error-500-following-upgrade-to-yuno-2-7/3395/6


alexAubin:
Merging this as a microdecision because this is an easy fix and a known bug for months

PR #99, opened 2018-05-09
Make overlay dragging work on mobile devices

nicofrand:
This MR uses touch events on mobile devices to allow to move the YNH overlay on mobiles too.
I am not proud of the code but making it better would probably require to rewrite the whole file :man_shrugging: (2018 style) .


Psycojoker:
Haven't tested but a quick check makes the code LGTM

Psycojoker:
Tested using chromium debug tools that simulate mobile behavior, works as expected.

For me we can merge tomorrow.

PR #102, opened 2018-06-24
[enh] Add PCRE regex support

zamentur:
As the original branch of this https://github.com/YunoHost/SSOwat/pull/44 is missing, I have create a new one.

## The problem
Lua pattern are not well known and a lot of packager don't know how write this kind of regex. PCRE seems more familiar.

## Solution
Add support of PCRE, and keep support of lua pattern during a transition period.

## PR Status
Tested with nextcloud

## How to test
./ynh-dev use-git ssowat
setup one of this app and check it works as expected in their install script:
- kodi, 
- nextcloud (feature using /%.well%-known/)
- lutim as private
- lstu as private
- bozon as private,
- jirafeau
- lufi
## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -188,6 +188,8 @@ end
 -- If the URL matches one of the `redirected_urls` in the configuration file,
 -- just redirect to the target URL/URI
 --
+-- A match function to support PCRE and lua pattern
+-- lua pattern will be deprecated in YunoHost
 function match(s, regex)
     if rex.match(s, regex) or string.match(s,regex) then
         return true
--- a/access.lua
+++ b/access.lua
@@ -188,14 +188,14 @@ end
 -- If the URL matches one of the `redirected_urls` in the configuration file,
 -- just redirect to the target URL/URI
 --
--- A match function to support PCRE and lua pattern
--- lua pattern will be deprecated in YunoHost
+-- A match function that uses PCRE regex as default
+-- If '%.' is found in the regex, we assume it's a LUA regex (legacy code)
 function match(s, regex)
     if not string.find(regex, '%%%.') then
         if rex.match(s, regex) then
             return true
         end
-    elseif string.match(s,regex)  then
+    elseif string.match(s,regex) then
         return true
     end
     return false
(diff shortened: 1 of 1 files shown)


Psycojoker:
I would be interested in having the feedback from the apps team as they are the first concerned, therefor pinging them cc @frju365 @JimboJoe @Josue-T @maniackcrudelis (or at the one I know to be actives)

Josue-T:
I had implemented the lua pattern in gogs and I confirm that it' a pain to use it. So I think that it's a really good idea to use PCRE instead.

Psycojoker:
@maniackcrudelis said on the chat that this was a good idea too and that it has already been discussed in the past apparently.

I'm personally a bit weirded out by this dual match approach, I hope this won't create weird bugs :/ Maybe we should have something like explicitely specifying the regex type? Like starting the regex string with "pcre:" and "luare:" (default)?

Otherwise, LGTM.

JimboJoe (approved):
This is very LGTM as well! :+1:

alexAubin:
Bumpitybump : what's the status of this ? Has it been tested ? Should it be merged ?

zamentur:
I suggest to put it in a testing release

alexAubin:
>Add support of PCRE, and keep support of lua pattern during a transition period.

Would there be a "simple" way to detect if the user is using Lua regexed or PCRE regexes ? So far it essentially looks like this will support both - but if we really want to push the usage to PCRE, we shall warn packagers about Lua regexes being deprecated somehow ?

This doesn't prevent from merging this though (I would be happy to merge this right away and include this in 3.3 testing)

alexAubin:
Discussing this with ljf today : there are not so many apps using those regexes in fact (c.f. list in original post) so it's not so important to check. But ljf will investigate a bit more the question of identifying lua regexes

alexAubin:
>But ljf will investigate a bit more the question of identifying lua regexes

Bump @zamentur ;)

(Otherwise we can also to just yolomerge this as it is, i dunno...)

zamentur:
Tested with the %.well%-known/acme%-challenge ssowat rules

zamentur:
I have found an error

PR #103, opened 2018-08-25
[fix] Secure cookie setting

frju365:
Need test... And I'm not LUA dev at all ;) (try to learn it)

Source : 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
One protection against CSRF BUT not supported by all browser... so need to be cumulate with other protections.

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -146,7 +146,9 @@ function set_auth_cookie(user, domain)
     local cookie_str = "; Domain=."..domain..
                        "; Path=/"..
                        "; Expires="..os.date("%a, %d %b %Y %X UTC;", expire)..
-                       "; Secure"
+                       "; Secure"..
+                       "; HttpOnly"..
+                       "; SameSite=Strict"
 
     ngx.header["Set-Cookie"] = {
         "SSOwAuthUser="..user..cookie_str,
@@ -165,7 +167,9 @@ function delete_cookie()
         local cookie_str = "; Domain=."..domain..
                            "; Path=/"..
                            "; Expires="..expired_time..
-                           "; Secure"
+                           "; Secure"..
+                           "; HttpOnly"..
+                           "; SameSite=Strict"
         ngx.header["Set-Cookie"] = {
             "SSOwAuthUser="..cookie_str,
             "SSOwAuthHash="..cookie_str,
@@ -180,7 +184,9 @@ function delete_redirect_cookie()
     local expired_time = "Thu, 01 Jan 1970 00:00:00 UTC;"
     local cookie_str = "; Path="..conf["portal_path"]..
                        "; Expires="..expired_time..
-                       "; Secure"
+                       "; Secure"..
+                       "; HttpOnly"..
+                       "; SameSite=Strict"
     ngx.header["Set-Cookie"] = "SSOwAuthRedirect=;" ..cookie_str
 end
 


helpers.lua
```diff
@@ -146,7 +146,9 @@ function set_auth_cookie(user, domain)
     local cookie_str = "; Domain=."..domain..
                        "; Path=/"..
                        "; Expires="..os.date("%a, %d %b %Y %X UTC;", expire)..
-                       "; Secure"
+                       "; Secure"..
+                       "; HttpOnly"..
+                       "; SameSite=Strict"
```
Psycojoker, line 151: This really needs testing as other websites being the sso might uses the cookie and that might break the whole SSO.

zamentur (commented):
I think this kind of things could be in a testing release, to evaluate possibility to make a version with it.

frju365:
Otherwise, i've begun to think (:P ) how we could implement CSRF protection in the form with token generation.

alexAubin (approved):
Soo, let's merge this in the next release with a testing before releasing as stable ?

I don't really know how to get feedback / spot issues otherwise ...

randomstuff:
AFAIU `SameSite=Lax` should be very fine.

randomstuff:
AFAIU, `SameSite=Strict` breaks links from one domain to another one (eg. links frome the YunoHost home page to SSOwat-protected apps installed in another domain).

alexAubin:
>AFAIU, SameSite=Strict breaks links from one domain to another one (eg. links frome the YunoHost home page to SSOwat-protected apps installed in another domain).

Thanks @randomstuff , I was indeed able to see that it breaks multi-site. Moved to `SameSite=Lax` in https://github.com/YunoHost/SSOwat/commit/7be6e76cb8fb60076e3cf711f027c3fa6e6938bf

frju365:
merci @alexAubin @randomstuff

randomstuff:
I turns out this does not prevent CSR on the admin interface. The `session.id` and `session.hashes` cookies are not affected (they are set by moulinette) and I can still CSRF the admin interface:

<pre>
HTTP/1.1 200 OK
Server: nginx/1.10.3
Date: Mon, 26 Nov 2018 00:35:30 GMT
Content-Type: text/html; charset=UTF-8
Content-Length: 9
Connection: keep-alive
X-SSO-WAT: You've just been SSOed
Set-Cookie: SSOwAuthRedirect=;; Path=/yunohost/sso/; Expires=Thu, 01 Jan 1970 00:00:00 UTC; Secure; HttpOnly; SameSite=Lax ;;
Access-Control-Allow-Origin: *
Set-Cookie: session.hashes="!Kt8mtVS9Mw+7vx+hiKoDPQ==?gAJVDnNlc3Npb24uaGFzaGVzcQF9cQJVB2RlZmF1bHRxA1goAAAAY2I3MzBhMjVlZTAwN2M0MzFiZjYxOGJmMTE4ZTljNTk2MzlkN2Q2MHEEc4ZxBS4="; secure
Set-Cookie: session.id=9d50d49ebce43ba0e864dd2f7eccca828178e95e; secure
Strict-Transport-Security: max-age=63072000; includeSubDomains; preload
Content-Security-Policy: upgrade-insecure-requests
Content-Security-Policy-Report-Only: default-src https: data: 'unsafe-inline' 'unsafe-eval'
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
X-Download-Options: noopen
X-Permitted-Cross-Domain-Policies: none
X-Frame-Options: SAMEORIGIN
Strict-Transport-Security: max-age=31536000;
</pre>

randomstuff:
It *does* block the [CSRF in vpnclient](https://github.com/labriqueinternet/vpnclient_ynh/issues/43).

frju365:
I didn't find cookie generation in Moulinette. An idea where it is ?

randomstuff:
It's in `_ActionsMapPlugin.login()` and `_ActionsMapPlugin.logout()` in `api.py`.

randomstuff:
[Bottle v0.12](https://bottlepy.org/docs/0.12/api.html) – the version shipped in stretch – does not support SameSite cookie. The [dev version](https://bottlepy.org/docs/dev/api.html) does.

PR #109, opened 2018-09-15
[mod] Redirect after logout if `r` URI argument exists

tituspijean:
### The problem
An app cannot ask for a logout without breaking UX: users end up on SSOwat login form, whereas going back to the app could be preferable in some cases.

### Solution
This PR addresses [YunoHost/issues#1201](https://github.com/YunoHost/issues/issues/1201).
The modification makes `logout()` helper test if `r` URI argument exists after logging out users. If so, the base64 address is decoded and nginx is tasked with the redirection.

### PR Status
Ready to be reviewed.

### Testing
Example: [Flarum package](https://github.com/YunoHost-Apps/flarum_ynh) is bundled with an SSOwat extension to support YNH users. It already puts an `r` argument while disconnecting YNH users and redirection has been successfully tested.

### Validation
- [ ] Quick review 0/1 :
- [ ] Simple test 0/1 :
- [ ] Deep review 0/1 :

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -889,8 +889,12 @@ function logout()
         flash("info", t("logged_out"))
     end
 
-    -- Redirect to portal anyway
-    return redirect(conf.portal_url)
+    -- Redirect with the `r` URI argument if it exists or redirect to portal
+    if args.r then
+        return redirect(ngx.decode_base64(args.r))
+    else
+        return redirect(conf.portal_url)
+    end
 end
 
 


alexAubin (approved):
Anyone has an opinion on this ? (including @YunoHost/apps ?)

Naively it looks legit to me but I haven't digged the issue / I don't have all use cases in mind

alexAubin:
Yolomerging because this looks legit~
Thanks !

helpers.lua
```diff
@@ -889,8 +889,12 @@ function logout()
         flash("info", t("logged_out"))
     end
 
-    -- Redirect to portal anyway
-    return redirect(conf.portal_url)
+    -- Redirect with the `r` URI argument if it exists or redirect to portal
+    if args.r then
```
zamentur, line 893: I think it should be in the "if is_logged_in() then" No ?

tituspijean: No, I don't think so. The function `logout()` already assumes that the user is logged in (as per line 884) and delete their cookies/session. Then my proposed code adds a check for the `r` argument to use it for redirection, and otherwise defaults to the initial redirection to portal.

PR #114, opened 2019-01-07
[fix] SSOwat crash after password change

Josue-T:
# Problem

SSOwat crash when we change the password.

I found the reason of this issue when I see the LDAP log. I saw that the LDAP connection fail here https://github.com/YunoHost/SSOwat/blob/stretch-unstable/helpers.lua#L353 by the error 49 (which is unauthorized).
I after saw that the connection was tried with the old password, which was invalid.

# Solution

Remove the password in the cash in the deconnection.

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -938,6 +938,7 @@ function logout()
         delete_cookie()
         cache:delete("session_"..authUser)
         cache:delete(authUser.."-"..conf["ldap_identifier"]) -- Ugly trick to reload cache
+        cache:delete(authUser.."-password")
         flash("info", t("logged_out"))
     end
 
--- a/helpers.lua
+++ b/helpers.lua
@@ -355,6 +355,15 @@ function set_headers(user)
             conf["ldap_identifier"].."=".. user ..","..conf["ldap_group"],
             cache:get(user.."-password")
         )
+        -- If the ldap connection fail (because the password was changed).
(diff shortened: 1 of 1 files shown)


alexAubin (commented):
That sounds legit to me but I really have a limited understanding of the whole cookie / ldap thing in ssowat.

Any opinion on this ?

alexAubin:
I tried the following (c.f. https://github.com/YunoHost/issues/issues/770 ) and it is not fixed by this branch I believe ... Can you clarify what exactly is the issue ?

I did : 
- Create a user
- Log in the SSO
- From the webadmin, change the password of the user
- On the SSO page, refresh
- Encounter error 500

(For now the manual fix to this is to *restart* nginx)

Josue-T:
Yes it's that issue that I encounter.

It worked on my side. But I will try a new time to test if this patch really fix the issue.

alexAubin:
Discussing with Bram : isn't it rather about changing the password from the SSO rather than the webadmin ?

Edit: well nope, I can't find/reproduce any issue related to changing password from the SSO ...

Josue-T:
Yes it's true. It's when you change by an other way than from the SSO. It's when I changed by the CLI that I saw this issue

Josue-T:
Now it should be fixed also when you don't disconnect before to change the password.

PR #117, opened 2019-02-01
Add themes support and simplify ynh inapps panel

eauchat:
Ok, I'll try to make it clear (sorry it's not short, but I tried to explain the change a couple of times, but I had the feeling it wasn't very understandable every time).

So. You know, the SSO homepage, the one that lists apps? You are redirected to it when you login successfully. But also, when you're in using an app, there is the yunohost button in the bottom right of your screen that let you toggle it's display.

So, before this PR, there was two codes to display this same content:
 (1) The first one is the page `info.html` (the one displayed when you visit /yunohost/sso/). It's display is generated by `info.html` + js from `global.js` + css from `ynh-style.css`
 (2) The second way (what you would see when you'd be using an app and clicking on the yunohost button (and it looks exactly like the first one)), was generated from `ynhpanel.js` and `ynhpanel.css`.

So the goal of this PR, is simply to remove the second "duplicate" code (2), and in `ynhpanel.js` to just call an iframe that loads the page `info.html`.

This allows to better separate functionalities:
 - info.html, global.js and ynh-style.css are where the display of the apps list page is decided,
 - ynhpanel.js and ynhpanel.css decide the way the "apps list panel" hides and shows on apps.

This PR is actually also doing a second little thing (sorry, I didn't manage to separate them, it seemed obvious to include it since the revamping of ynhpanel.js was meant for this purpose), it's to add compatibility with the under creation themes API, so the SSO can display custom themes.

Hope that's understandable :-)  :heart:


alexAubin:

PR #118, opened 2019-02-01
Some example themes to test

eauchat:
This PR just contains all changes from #117 plus some basic example themes to easily test the new "yunohost themes" API.


portal/header.ms
```diff
@@ -34,10 +37,6 @@
 </head>
 <body class="{{#connected}}logged{{/connected}}">
 
-  <h1 id="logo" class="logo">
-      <img src="assets/img/logo-ynh-white.svg"/><span class="element-invisible">Yunohost</span>
-  </h1>
```
alexAubin, line 39: Why remove this tho ? D:

portal/assets/js/global.js
```diff
@@ -35,4 +68,20 @@ document.addEventListener('DOMContentLoaded', function() {
     addMaildrop.parentNode.insertBefore(inputDropClone, addMaildrop);
   });
 
+
+  // Get user's infos
+  var r = new XMLHttpRequest();
+  r.open("GET", "/ynhpanel.json", true);
```
alexAubin, line 74: Eh, where is that supposed to come from ? x_x

alexAubin: Hm now that I dig a bit deeper I kinda understand, but while testing on my machine, before logging in, it can't access this file, and after logging in there's no 'theme' property yet defined apparently :s

portal/header.ms
eauchat: Hey, yes we discussed a bit about this in #121 with @lfuelling.
I had chosen to remove the logo here because it was confusing to have a logo that's clickable in the panel and not clickable in the homepage.

In the end, IMHO the best option would be to have a logo that's always there and a button that's clickable, so yes, maybe not remove those lines (and also the styles necessary for it).

portal/assets/js/global.js
eauchat: Well this comes initially from the ynhpanel.js file, it was used to fetch config and is not necessary there anymore. I can't say I really understand the organization behind the serving of the config, so I decided to use the same system that was used and worked.

eauchat: > there's no 'theme' property yet defined apparently

ah yes I guess if you've never defined the theme from the command line with the other PR in yunohost-admin it will not be defined I guess

eauchat: > before logging in

I didn't test pre-logging, true. But yes it would be nice that the theme extends to the login page. Maybe we can figure out another way to pass the chosen theme setting, but this is the part of code that I understand less.

portal/header.ms
lfuelling: #121 reverts this change and also fixes another issue with the overlay (see my review in #117).

My approach to the logo would be to disable it in a theme when it's not wanted, but maybe this also could be done using configuration (a bit overkill imo).

alexAubin:

PR #120, opened 2019-02-12
add missing dependencies to README

lfuelling:
Hi,
I found missing dependencies while trying to get this to run.

PR #121, opened 2019-02-12
Theming support (includes vaporwave theme)

lfuelling:
Hi,

as I announced on [Mastodon](https://comm.network/@lerk/101580319862703664), here's the PR including theme support (using the config file, not yet per-user) and my custom vaporwave theme inspired by the theme of my Mastodon instance.

Unfortunately this includes #120 because I was stupid while creating the branch.

Here are some screenshots (believe me, the gradient looks better in [real life](https://lewd.club)):

login view

app view

**EDIT:** Sorry, I saw #117 too late. I rebased onto that and made some fixes. (@eauchat it would be nice if you could give feedback on whether I got the changes you intended right)


lfuelling:
I also changed the behavior of the user menu (on the top left).
When in the overlay (iframe), clicking the menu will navigate back to `/sso/info` so paranoid people like me can be sure there is nothing else running on there site where they enter their password.
This might be too paranoid for official packages, but since users are able to run basically *anything* with this overlay showing, it seemed like a good idea. People who are not that paranoid still can click "edit my profile" in the footer and will be taken to `sso/edit` inside the iframe.

lfuelling:
Please don't merge this right now. I found a bug that when the iframe was opened once, it doesn't get closed properly (more like hidden) and the user is unable to click anything except the (now hidden) links on the overlay.

I will look into this when I get home.

eauchat:
Hey, I'm just discovering this now. I'll try to have a look a the changes you made as soon as I make time, but it's cool to have other people working on theming :)

About the ynh logo, I had intentionally removed it, so it appears only in the panel display (= the iframe) because I found it a bit confusing sometimes to have it in the home page, it makes you want to click on it but it's not clickable. But maybe it's also good to have it.
Or maybe it's possible to distinguish between a navigation button (the one you have in the iframe) and a logo that is always present and not clickable.
Maybe that'd be more straightforward, I don't know. What do you think?

lfuelling:
I think it's pretty clear that the button (which has a black background and is dragable) is something else than the logo which is always at the same position.

But I also tried clicking it multiple times. :grin: 

I would leave it because it looks more consistent with the logo than without.

alexAubin:

PR #116, opened 2019-01-28
First tests to introducing custom themings.

eauchat:
Along with pull request in "yunohost" repo (that will be cross referenced bellow), this is implementing the necessary bits to modify SSO display depending on the chosen custom theme.
I joined two ultra basic custom themes for the sake of the demonstration.

Note: I could work on making one or two proper themes that could be included with ynh by default on install, in addition to the possibility to add any new custom ones.

PR #122, opened 2019-02-21
Theming reloaded

alexAubin:
https://hooktube.com/watch?v=FDkhqjvS43E

Sooo I tried to merge the various work from #116, #117 , #118 and #121 because it was quite confusing because there are many different things to handle (and the specs of theming were not so well-defined) ... There are also various things which I am not sure to understand, mostly related to the historical code for SSOwat...

TODO : 
- [x] check / include chateau theme example
- [x] have a look at all the FIXME's
- [x] test in various case (logged in / not logged in, check the 'yunohost' button and overlay, ...)
- [x] Try to remove some JS madness
- [x] Implement a nice "light" theme (in opposition to the default 'dark' theme)
- [x] Write some documentation on how to add / build a new theme
- [x] There is a weird issue when logging-out from the 'overlay' thing : you don't truly get redirected to the actual login form, you stay in the overlay ... Probably related to the fact that it's done in an iframe
- [x] Make it easy to change the yunohost logo by something else
- [x] Make it easy to change the yunohost background by something else
- [x] We might want to tweak `yunohost_panel.conf.inc` in core instead of loading 3 different piece of js/css from the JS, otherwise that's a huge mess ...
- [x] Maybe we should split custom.css (from themes) into a custom_portal.css and custom_ynhicon.css ... to be discussed

Test cases to test / validate : 
- [ ] Login page
- [ ] Portal page
- [ ] App page when it include ynhpanel.js (the icon in the corner)
- [ ] The portal overlay when clicking on the icon in the corner when being in an app
- [ ] All of those both for the default theme and the example themes ...
- [ ] Annnnnd check that everything looks decent on mobile ...


portal/assets/js/global.js
```diff
@@ -68,4 +68,21 @@ document.addEventListener('DOMContentLoaded', function() {
     addMaildrop.parentNode.insertBefore(inputDropClone, addMaildrop);
   });
 
+  // FIXME - I don't understand what this do ...
+  // This looks kinda hackish :|
+  if(window.location != window.parent.location) {
```
lfuelling: This condition is true, when the page is inside the iframe (page url != html url)

portal/assets/js/global.js
```diff
@@ -68,4 +68,21 @@ document.addEventListener('DOMContentLoaded', function() {
...
+    // Set class to body to show we're in overlay
+    document.body.classList.add('overlay');
+      let userContainer = document.querySelector('a.user-container');
+      userContainer.classList.replace('user-container-info', 'user-container-edit');
+      userContainer.setAttribute('href', userContainer
```
lfuelling: This changes the `href` from `[...]/edit.html` to `[...]/` (which becomes `[...]/info.html` eg. the main page)

portal/assets/js/global.js
```diff
@@ -68,4 +68,21 @@ document.addEventListener('DOMContentLoaded', function() {
...
+          .getAttribute('href')
+          .replace('edit.html', ''));
+      userContainer.addEventListener('click', function(e) {
```
lfuelling: This prevents the link's default behavior (which is navigate inside the iframe) and navigates to the page on the main page instead of inside the iframe.

lfuelling:
Hey @alexAubin, 
I added some comments to the sections I wrote. Hopefully it clarifies their purpose.

portal/assets/js/global.js
```diff
@@ -68,4 +68,21 @@ document.addEventListener('DOMContentLoaded', function() {
     addMaildrop.parentNode.insertBefore(inputDropClone, addMaildrop);
   });
 
+  // FIXME - I don't understand what this do ...
```
lfuelling: Basically this is the "feature" I described in [this comment](https://github.com/YunoHost/SSOwat/pull/121#issuecomment-463161831).

.gitignore
```diff
@@ -1,4 +0,0 @@
-*.swp
```
lfuelling, line 1: The `.swp` ignore was  there before I added the other stuff.

portal/assets/js/global.js
```diff
@@ -68,20 +68,4 @@ document.addEventListener('DOMContentLoaded', function() {
...
-  var r = new XMLHttpRequest();
-  r.open("GET", "/ynhpanel.json", true);
-  r.onreadystatechange = function () {
-    // Die if error
-    if (r.readyState != 4 || r.status != 200) return;
-    // Response is JSON
-    response = JSON.parse(r.responseText);
-
-    // load additional theme's script
-    loadScript("/yunohost/sso/assets/themes/"+ response.theme +"/js/ynhpanel.js");
```
lfuelling: I'd also rather not have themes load js code. CSS in enough IMHO.

helpers.lua
```diff
@@ -586,7 +586,7 @@ function get_data_for(view)
     data['flash_fail'] = {flashs["fail"]}
     data['flash_win']  = {flashs["win"] }
     data['flash_info'] = {flashs["info"]}
-    data['theme_name'] = ""..conf["theme"]
```
lfuelling: Sorry, that was a debugging measure I forgot to change back. (I don't know lua at all so I had to try some stuff)

.gitignore
alexAubin: Yes, don't worry :wink: 
I was just a bit frustrated that this .gitignore was just getting filled with various tmp files coming from stuff which have basically nothing to do with ssowat itself and i don't think we want to have 10+ lines filled with that kind of stuff. Anyway that's not a big deal but I propose to just get rid of the .gitignore

helpers.lua
lfuelling: Also it* didn't work for me *at all* without the stuff that I consider to be a typecast to String (`""..var`).

*it => getting the theme name

portal/assets/js/global.js
alexAubin: Ah yes, in fact I don't know about this because I didn't do so much theming in my life. If you think JS ain't needed then that's fine with me ?

portal/assets/js/global.js
alexAubin: Wow that's pretty interesting, wouldn't have thought about that kind of stuff o.O" Now you make me wonder if we shouldn't just do this as well for the "edit my profile" at the bottom as you mention it :P

portal/footer.ms
```diff
@@ -11,5 +11,8 @@
 
   <!-- Scripts -->
   <script src="assets/js/global.js"></script>
+  {{#theme}}
+  <script src="assets/themes/{{theme}}/global.js"></script>
```
lfuelling: So every theme has to implement Javascript or the request for this element will fail. (Which is only visible when the devtools are open but imo it's a bad practice.) 

Would it be possible to detect if a js file exists for given theme and only put in the script tag if it does?

portal/assets/js/global.js
lfuelling: I thought about that too.

.gitignore
lfuelling: Yes, I can also put all the stuff into my global ignore file. (Which would be best practice anyway :p)

portal/assets/js/global.js
lfuelling: No, not really. Most stuff can also be done using CSS.

helpers.lua
alexAubin: Uh that's weird o.O It seems to be working on my side ?

lfuelling: Okay, Ignore my comment then :grin:

portal/assets/js/global.js
eauchat: Personally, I think it would really good to have the possibility to run js in themes. If you want to do some real fun things, you will need it.

alexAubin: Hmyea naively I too think that brings more flexibility ...

lfuelling: It does. But it's a security hole. 
(Considering it might be possible in the future to 1-click-install random themes)

alexAubin: Kind of yes, but I don't think we would allow to install themes which did not have some sort of reviewing from the community like we do with the community app list... So, yes, can be seen as a security hole, but not really much a bigger hole than installing community apps in the first place (which run bash scripts as root) :P

eauchat: I guess but that's why there would be official and non official themes (like for app) and a warning if you want to install external themes, no?
You have to trust someone at some point, when you install something.

eauchat: IMHO not letting js is keeping the power to do nice things only in the hands of YNH devs and to me, one cool thing of FOSS is when they make it easy for more people to do power stuff.
After, I don't know the whole extent of security issues, I just know that the first theme I made for YNH needed javascript to be achievable ;)

lfuelling: Okay, I'm outvoted I guess :grin: 

Please take into account my comment to the other location where theme js is loaded, regarding the 404 that will occur with themes that have no js.

alexAubin: I think it really boils down to what we allow in terms of "installing custom themes". For now I think we should do the following : 
- ship a few "official" themes
- allow people to easily add their own theme / tweakins, e.g. copying one of the default themes and tweaking the css and so on
- if people want to share theme with each other, they should wget and extract some zip archive (and of course realize that they are running other people's code on their machine ...) so still doable, but that requires more than 1-click.

In the long term : 
- We can allow people to change theme from the webadmin
- Allow people to import a new theme somehow (idk, maybe entering an url to a zip archive) and have a warning explaining that they should be careful ... Maybe we can have a repo with themes, and people would be able to submit PR to this repo to add their own theme to be reviewed by the community, idk. We have time to think about this.

But the bottom line is that we shall indeed avoid 1-click install where people ~~accept candies from strangers~~install themes from random third-parties without being aware of possible implications

eauchat: > Please take into account my comment to the other location where theme js is loaded, regarding the 404 that will occur with themes that have no js.

Yep, I also saw this issue when coding, but I didn't know a simple way to solve it. Checking that a file exist isn't so straightforward with js, and in the end I decided that it's not such a big deal if it just doesn't file the file. It's the same for styles, it just warn a 404 but doesn't really create an issue I think.
Another way would be to specify in the doc that it's good practice to have those files, even if they are empty.
But all that is not so optimized, I agree it would be cool to be able to make the request only if the file exists.

lfuelling: > > Please take into account my comment to the other location where theme js is loaded, regarding the 404 that will occur with themes that have no js.
> 
> Yep, I also saw this issue when coding, but I didn't know a simple way to solve it. Checking that a file exist isn't so straightforward with js, and in the end I decided that it's not such a big deal if it just doesn't file the file. It's the same for styles, it just warn a 404 but doesn't really create an issue I think.
> Another way would be to specify in the doc that it's good practice to have those files, even if they are empty.
> But all that is not so optimized, I agree it would be cool to be able to make the request only if the file exists.

Yes, I'd try checking for the file in lua and printing some boolean value corresponding to this check into the js code, which in turn would stop the `script`-tag to be added when no js file was found.

lfuelling: > But the bottom line is that we shall indeed avoid 1-click install where people ~~accept candies from strangers~~install themes from random third-parties without being aware of possible implications

:grin: 
If no 1-click install was available, the user can be blamed enough for putting malicious code into the application (by themselves).

portal/assets/js/global.js
```diff
@@ -62,6 +39,26 @@ document.addEventListener('DOMContentLoaded', function() {
...
+    for (var i = 0; i < app_tiles.length; i++) {
+        app_tiles[i].addEventListener('click', function(event) {
+        // if asked to open in new tab
+        if (event.ctrlKey || event.shiftKey || event.metaKey
+            || (event.button && event.button == 1)) {
+            return
+        }
+        // if asked in current tab
+        else {
+            event.preventDefault();
```
lfuelling: I would add a call to `event.stopPropagation` to prevent the event from bubbling up.

portal/assets/js/global.js
eauchat: @alexAubin I agree with all your points, I think the admin theme install part should be designed like the one for apps (which should/will be changed a bit I guess YunoHost/issues#1302).
Official themes, easy to install in one click, non official with warnings, and make it easy to customize (duplicate, modify) themes.
I think a repo with themes is a cool idea.

lfuelling: A repo with themes would be awesome!

It would be even more awesome if it was a submodule of SSOwat but idk if you're up for the hassle this brings :)

alexAubin:
Sooo I fixed quite a lot of JS madness .. I ended up merging global.js into ynhpanel.js because I did not understand the point of having two separate files ... Hopefully it should be much easier to understand what's going on, starting here : https://github.com/YunoHost/SSOwat/blob/ad27933f5485ae6fd23b6f2d3924185564180f63/portal/assets/js/ynhpanel.js#L201-L220

Needless to say I would never have been able to do this without eauchat's initial refactoring.

I guess a similar work could be done for ynhpanel.css vs. ynh-style.css but I'm not comfortable enough with css :sweat_smile:

lfuelling:
I think it's better not to merge those files to keep `ynh-style.css` "small" (< 800 lines) :)

alexAubin:
Weeeell, obviously I would agree, but why should one be small and the other one big ? At this point, ynh-style is about 800 lines and ynhpanel.css is 165 lines ... But my understanding is that they are basically both loaded on every page behind a yunohost (one via the ynhpanel.js, and the other from the portal itself, loaded in the overlay) ?

lfuelling:
`ynhpanel.css` is not loaded on `/yunohost/sso/`. The stuff inside seems to affect only the overlay and ynh-button.

Maybe some things could be ported from `ynh-style.css` to `ynhpanel.css` if they're only used there.

alexAubin:
Sure, but the bigger one, `ynh-style.css`, is included in the portal, and the portal is included in everypage in the overlay ... So if we are worried about performance, that's what we should be pay attention to ?

lfuelling:
I was more worried about maintainability in this case. 

From a performance point of view it might be wise to start using a minifier/preprocessor. 
Although I can't think of any better time to do the preprocessing than as a precommit hook in combination with something like a `/dist` folder.

Psycojoker:
We are already using gulp for the admin for minification, although I really don't like it, it would be easy to port it for SSOwat since our toolchain is already adapted for that.

eauchat:
Hey, I can't have look at the code right now but whenever I can I will, so I understand better this idea of merging global.js and ynhpanel.js.
Originally those two files where separate and doing roughly the same things, each independently from each other.
For me the point on keeping two files was to separate what's about the portal and what's about the panel (= the fact that the portal can be shown/hidden from an app).
Same thing for css.
I think even a bit more fragmentation would be better in fact: maybe a global.js-global.css loaded in all SSO pages, and a loggedin.js for what's happening in logged in pages, or maybe better one js per page?? (portal.js, edit.js...), maybe same for styles. And a panel.js-panel.css for what concerns the display of the portal from apps.

I feel it would be much simpler to have small files loaded only when they are useful, would be one or two max per page, but different ones depending on the page. (Maybe it could remove the need to minify anything.)

I'm thinking out loud, without looking at the code. But...
What do you think?

I really want to help on this, but these days are not so good for me. But if I succeed making some time I will try to :)

alexAubin:
Well I really thing we don't need to manify anything so far .. The total weight of the JS file is about 12 ko which is fine I think. Then there's the question of wether or not to optimize the number of requests (which is opposite of fragmenting the code in multiple files ;)). In fact right now I'm worried about the fact that loading the two regular CSS files + the regular JS + the custom CSS from theme + the custom JS from theme **on every app page** is already too much, so I would try to avoid fragmenting even more ...

Anyway, I merged those two global and ynhpanel.js files : 
- for semantic, because it was really not clear what piece of code was meant to use and in which context ... I tried to make it much clearer
- for style consistency : for example in the first file, the "DOM loaded" event was handled one way, and in a completely different way in the second file... Also ynhpanel.js adds some helper which are useful to clarify the code of global.js (instead of absolute madness like `foo && [].forEach(foo, function() { ...`)

eauchat:
Maybe I'm making it too complicated, and having so many files is just useless in fact.
I was just thinking of the issue you were mentioning Aleks of having js code that concerns the display of the apps list in a js that is loaded in the login page.
It would be nice that the code that concerns the listing of apps is just loaded in the page where it's needed.

alexAubin:
Yes indeed, but working more on this I think it's really no big deal. Basically the difference is that it's gonna perform a loop on an empty array and it's not so much of a big deal ... I was just confused by so many different things that this detail struck me at that time but now it's a bit clearer in my mind (though there definitely still are weird stuff in there I suppose :D )

If we really want to, I noticed that the `<body>` has the class `logged` when user is logged in so we can add an `if (! logged) {return;}` somewhere to avoid attempting to run irrelevant code but that's just a detail I think

eauchat:
Ok, for me the idea was more that it's not necessarily so cool to send some code about what's happening when you're logged in, when you are not.
I think it would be wiser if you cannot know what's happening inside when you're logged out.

alexAubin:
Well really what's happening is that it's simply trying to grab a few elements based on their class / ids and change some styling and stuff, so as said it's really basically doing a look on an empty array ...

Then if you are concerned about security : yes, ideally not having JS involved would be nice. And in our case, this JS ain't necessary to log in so it could be remove. But in the future (or in a theme using fancy stuff) it could be ... But security-wise, also consider that if you don't trust the website you are logging in, you should not attempt to log in in the first place :/ We can't really do much about that..?

alexAubin:
Ah I think I understand what you mean now ... But well I don't really think it's a big deal... it's not as if it was supposed to be secret. In fact, license-wise, YunoHost is AGPL, and per the clause of AGPL, even if you do customize some stuff (like the theming and JS and so on), you are supposed to make your modification public (even if they are just meant for you).

lfuelling:
No security relevant code should be run in client js anyway. That part should be left to the server as any client can edit the code.

Regarding the file size: I wouldn't worry about that, as Nginx has caching capabilities and the css/js *should* only be loaded when it changed or on the first visit to the page*.

*: I don't know if that's switched on by default in ynh

alexAubin:
I think it's not enabled ;D : 

```
# grep -nr "cache" /etc/nginx
/etc/nginx/conf.d/yunohost_admin.conf:21:    ssl_session_cache shared:SSL:50m;
/etc/nginx/conf.d/yunohost_admin.conf.inc:9:    # Short cache on handlebars templates
/etc/nginx/conf.d/ssowat.conf:1:lua_shared_dict cache 10m;
/etc/nginx/conf.d/yolo.test.conf:30:    ssl_session_cache shared:SSL:50m;
```

ssl_session_cache is only about the SSL part I guess

eauchat:
I'm not really getting what are the three options [here](https://github.com/YunoHost/SSOwat/blob/32323b49808982341ff3b6a18e82cf4c65c20f17/portal/assets/js/ynhpanel.js#L203-L206).
From what I was understanding, there was two cases:
  - loading directly from visiting `yunohost/sso`,
  - loading from an iframe in an app (and that iframe simply shows `yunohost/sso`).

What's the third one?

eauchat:
Oh, I guess I understand. This is because ynhpanel.js is loaded both in `yunohost/sso` and in the apps.
I guess your goal doing that was to be able to share helpers useful in both codes.

alexAubin:
Yes, it's loaded in the apps themselves (in that case, the relevant JS is for instance the code that manages the display and dragging of the small "panel button")

>I guess your goal doing that was to be able to share helpers useful in both codes.

Yes, but mostly it was just too confusing to understand what piece of code was loaded and when ... `ynhpanel.js` was loaded all the time in everypage (like now) and then *inside* the portal, you would get an additional `global.js` which does a few things related to the portal, and the two JS files were having different coding style which is just complete madness. Considering `global.js` was small, one might just as well merge it in `ynhpanel.js` and make everything simpler to understand.

eauchat:
Thinking of the ynh logo/portalButton, since it would be nice to be able to setup any file+filetype you want for it (png, svg)...
What do you think of serving all files contained in the current themes folder? I guess this must be doable from access.lua.
You could then query all the current theme related files without needing a specific rule for each file: `/ynhtheme/custom.js`, `/ynhtheme/custom.css`, `/ynhtheme/background.jpg`, `/ynhtheme/myCustomLogo.svg`.

lfuelling:
But wouldn't this also require every theme to use those files (or bundle empty/"default" files)?

eauchat:
Mmmh maybe my comment wasn't very clear. I can try to rephrase it.

The idea would be that the SSO would serve any file contained in the themes directory.
It means for example that from your theme you can specify css rule like this: 
```
background-image: url("/ynhtheme/mynicelogo.svg")
```
And if you have this `mynicelogo.svg` file in your theme's folder it will be served.

It's doesn't address the issue of calling for absent themes custom.js/custom.css. It's to bring possibility to have custom logos or other medias.

eauchat:
Hey, nice for the merging.
Funnily I was just working on making content from the themes directory available to root like I said.
I didn't know a good approach so I tried making a basic ls of the directory, but I didn't get to make it work so far I don't understand why it just doesn't work to serve files, even if the ls works. I guess it could be some kind of lua specific behavior that I don't know of.
I'll try some more tomorrow.

eauchat:
So... the last PR I posted, should solve:
- [X] include some theme examples
- [X] have a look at the FIXMEs (there was only one left, and personally I had no issue about the portal button not being dragged correctly if moving mouse too fast, for me everything behaves fine)
- [ ] Implement a nice "light" theme (I don't know exactly what you meant by this one, so don't know if it's done ;)
- [ ] Write some documentation on how to add / build a new theme (I can try to work on this next)
- [X] Make it easy to change the yunohost logo by something else ?

Questions:
I don't think everything has to be perfect to start kicking in the theming feature, but with the current styles, it requires quite a lot of css to simply change the background of the page for example.
Made me wonder if it would not be nice to simplify even more the css.
Just an idea. Interested in other's opinion, I wouldn't mind working on it, if other people find that useful too.

alexAubin:
>I don't think everything has to be perfect to start kicking in the theming feature, but with the current styles, it requires quite a lot of css to simply change the background of the page for example.
Made me wonder if it would not be nice to simplify even more the css.
Just an idea. Interested in other's opinion, I wouldn't mind working on it, if other people find that useful too.

Yes :+1: To me that sounds like a good idea

Also thanks for all the mentionned work :wink:

alexAubin:
I'm currently doing some test and the CSS for theme is a bit of a mess ... We need to clarify some stuff because currently you can't really change `body`'s CSS (or any "standard" stuff like `a`, `table`, ...) because `custom.css` will be loaded in all apps and that might change the CSS of every app ...

So either we should : 
- split the stuff into two css files, one only to customize the display of the "portal icon", the other one to customize the overlay
- or keep a single file, but prefix ~everything with ynh- to avoid conflict with the original CSS of apps ...

eauchat:
@alexAubin Hey just seeing your comment (didn't check your changes yet so I don't know what you chose).
But I strongly vote for dividing css for different uses. The ones for the ynhpanel, the ones for the portal.
But I guess you already knew what would be my opinion, because I was saying the same for js code before ;)
To me this makes things easier to code and to understand. Now I know the current logic, but sometimes even knowing it, I forget that js and css are loaded in all pages and have issues when trying to create themes and I created lots of problems.
Like what you mention of styling the body that will change the style of every app.

eauchat:
I saw afterwards the playings you had with themes and looking forward to testing changes you made ;)

alexAubin:
Sooooo let me try to summarize the various things I've been working on : 

- split the CSS into two specific files : portal and overlay - so that you don't have to make sure that you rules for the portal won't cause any trouble on apps ...
- however I kept a single JS files ... partly because of laziness and also to avoid an additional request / file to handle ... also because the portal and overlay kinda use the same helpers so splitting the file would mean duplicating code
- implemented a light theme (in opposition to the default "dark" theme) : c.f. screenshot of [login](https://framapic.org/Op2jzTqQQzvm/mlWdp9qqIsu0.png) and [app list](https://framapic.org/H2zKTQZiUQma/RQtvAgp0UK3k.png) - inspired by what I saw a while ago on Open Tunisia
- removed some JS madness loading the additional CSS and JS after DOMready event ... c.f. https://github.com/YunoHost/yunohost/pull/689
- fixed an issue with dragging of the small yunohost button/tile inside the overlay iframe
- enabled a 1-hour cache on all the portal assets to reduce the number of request ... 
    - I choose 1 hour because some of those assets may change (e.g. if the admin changes the theme or just the logo .. you don't want to have to tell all your user to force a cache refresh ... I guess 1 hour is an acceptable tradeoff)
- many attempts to improve the semantic of various stuff, + various small tweaks / bugfixes
- documented the whole thing, c.f. : https://github.com/YunoHost/doc/pull/957

Soooo to me the work is **complete**. The last thing I'm wondering about is what themes to really include. Imho the `light` theme is definitely to be kept (apparently we tend to underestimate how dark interface make people afraid...). The `clouds` and `vapor` theme are kinda fun and show example of what can be done. However imho the `random` theme imho is a bit too much... so I propose to remove it (or it should be optimized to avoid copypasting n time the logo just to change the color :s ).

eauchat:
Hey, some quick remarks...
It's really cool all the work :)
I totally agree with you that it's not necessary to include the random theme by default. It needs improvement, and I think it's place is more in a repo somewhere with custom themes that people can install if they want.
One proposal: let's rename the info.html page to portal.html, would be more clear no?
And it's been a while that I was thinking it would be also cool somewhere for admins to actually be able to set a completely custom portal page. In other words, instead of /yunohost/sso/ serving info.html, it would serve whichever page the admin decided/created instead.
I think it would be really cool, and I imagine it might not be so complicated to do. But maybe it's not actually something to do in sso repo but in yunohost repo?

alexAubin:
Planning to add this to the 3.5.x which I would like to release early next week ... so this should be merged in the next few days

PR #123, opened 2019-02-26
Simplify tiles styling

eauchat:
Having the name of the app as data attribute of tiles makes it much easier to (custom) style them.


eauchat:
Also, the styling of tiles IMHO belongs to the chosen style, so it should happen in custom.js. As a matter of fact, vaporwave style doesn't need the default styling (I simplified vaporwave for that).

alexAubin:
Yolomergin', cheers !

PR #124, opened 2019-03-09
Make current theme's files accessibles under "/ynhtheme/"

eauchat:
This commit makes any file present in the current theme's directory to be served at "https://server.tld/ynhtheme/path/to/file.ext" (yes it works recursively and only serve files).

For example let's say I have a custom logo located in `portal/assets/themes/myCustomTheme/customlogo.png`. This logo will be accessible from "https://server.tld/ynhtheme/customlogo.png".
So then I just reference it from my `custom.css` file with:
```
#logo, #ynh-overlay-switch {
  background-image: url("/ynhtheme/customlogo.png");
}
```


eauchat:
I made little helpers to make it more convenient to serve custom paths (serveAsset). There's also a helper to execute a callback on all files in a directory (scandir).
I left them in `access.lua`, but maybe those helpers could be put in `helpers.lua`.
I don't know what's best.

eauchat:
Two new themes to come, making use of these functionalities, with custom logo... :-)

eauchat:
There is still (at least) one issue to the two themes: the custom ynh logo is not served when logged out.
Still it's important to be able to customize the login page.
Didn't look yet how to make sso serve those files as well. Any suggestion welcome :)

PR #126, opened 2019-03-14
Vapor95

alexAubin:
I let myself go with the vapor theme and this happened ... Dunno what to do with this, I probably went too far away from the original spirit '-'

Also the CSS is a mess

2019-03-14-223803_1366x768_scrot
2019-03-14-223827_1366x768_scrot
2019-03-14-223846_1366x768_scrot
2019-03-14-223916_1366x768_scrot


lfuelling:
I like the idea!
But maybe make it an entirely new theme and use another background.

I'm thinking about something like this, but more pixelated:
space background

And imo the alerts should be marquees :D

PR #131, opened 2019-03-20
Theming reloaded smallfixes

eauchat:
In portal, scroll .content, not body (otherwise if a custom background image is set the background is also scrolling).
I made it global, not only in clouds theme, because it makes customization more straightforward.
Also had to change a bit the css rules for body, otherwise it fucks the login page.


alexAubin:
Tested and works very well, thanks !

PR #132, opened 2019-03-20
Theming reloaded redirections improved

eauchat:
With this commit, if you query /yunohost/sso/edit.html page when you're logged out, it will redirect you to login page, then if you login successfully redirect you to the edit page.


alexAubin:
Yolomergin'

PR #133, opened 2019-03-22
Support for userinfo json

zamentur:
## The problem
Before there was ynhpanel.json we could use to get info about apps the user can use. It could be useful to display this kind of info on a page, like menu_ynh or other things like that.

## The solution
Restore this feature, but rename the file to explain better the goal of this file.

## Status
Untested

## Validation

- [ ] Quick review 0/1 : 
- [ ] Simple test 0/1 : 
- [ ] Deep review 0/1 :


portal/assets/js/ynh_userinfo.json
```diff
@@ -0,0 +1 @@
+{}
```
alexAubin, line 1: I still don't understand why we need this empty file :/ Can't we just return an empty dict when needed with creating this file x_x (I'm not even sure to understand how/if that file is actually likely to be used somewhere)

access.lua
```diff
@@ -334,6 +334,7 @@ end
 
 if hlp.is_logged_in() then
     -- serve ynhpanel files
+    serveAsset("/ynh_userinfo.json", "js/ynh_userinfo.js")
```
alexAubin, line 337: This doesn't work ... Accessing domain.tld/ynh_userinfo.json doesn't return the expected info... Also this function `serveAsset` will then apply public-caching ... but since this json is meant to contain private info, this should probably be cached as private instead

alexAubin (changes requested):

alexAubin:
Closing due to lack of activity and everything works without this

This will be addressed in the upcoming portal API

PR #136, opened 2019-05-17
Allow access in portail in other domain than main domain

Josue-T:
# Problem:

https://github.com/YunoHost/issues/issues/1335

# Solution

Allow to load the portail in other domain when the user is logged in.

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -65,7 +65,7 @@ end
 -- If the URL matches the portal URL, serve a portal file or proceed to a
 -- portal operation
 --
-if ngx.var.host == conf["portal_domain"]
+if (ngx.var.host == conf["portal_domain"] or hlp.is_logged_in())
    and hlp.string.starts(ngx.var.uri, string.sub(conf["portal_path"], 1, -2))
 then
 


opi:
Yeah, I guess you found it !

I removed my previous hacks (on ynh_portal.js and the nginx/iframe part), and your solution works fine ! 

Now when I'm logged in, https://sub.domain.tld/yunohost/sso/ brings me the portal, and when I'm logged out it returns a nice 404. Perfect !

opi (approved):

alexAubin:
That sounds sort of legit to me
Though I'm really wondering how that `is_logged_in` helper works in cross-domain context x_X
Will test and think about it a bit more

opi:
> Though I'm really wondering how that is_logged_in helper works in cross-domain context 

Because the SSOWat auth cookie is explicitly set with `Domain=.domain.tld` (Source: https://github.com/YunoHost/SSOwat/blob/a194168b84a85fe2e7ff3cbc5d6619910b9f6b49/helpers.lua#L149)

https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Scope_of_cookies
"Domain specifies allowed hosts to receive the cookie. If unspecified, it defaults to the host of the current document location, excluding subdomains. If Domain is specified, then subdomains are always included."

alexAubin:
Interesting :+1: 

So I think it covers the case maindomain=`foo.tld` and subdomain=`sub.foo.tld`

But I doubt it covers the case maindomain=`foo.tld` and subdomain=`bar.tld`

(Also not sure about maindomain=`portal.foo.tld` and subdomain=`sub.foo.tld`)

opi:
> But I doubt it covers the case maindomain=`foo.tld` and subdomain=`bar.tld`
> 
> (Also not sure about maindomain=`portal.foo.tld` and subdomain=`sub.foo.tld`)

Never try these case, but does the portal button even work ?

Josue-T:
I need to say that this patch work only with the `protected_url`. So for all `protected_url` you need to be authenticated so it's the reason why it work.
And to understand why the cross domain authentication work the code is here : 
https://github.com/YunoHost/SSOwat/blob/stretch-unstable/access.lua#L37-L57
https://github.com/YunoHost/SSOwat/blob/stretch-unstable/access.lua#L128-L140

I'm also working an other patch which as the same target than #119 but with an other way wish should work better.

alexAubin:
Yolomergin for 3.6 testing

PR #137, opened 2019-05-23
Prevent duplicate portal button when app uses iframes. Fix #1354

opi:


opi:
`window.frameElement` doc and browser support https://developer.mozilla.org/en-US/docs/Web/API/Window/frameElement

decentral1se (approved):

alexAubin (approved):

Psycojoker:
Related issue https://github.com/YunoHost/issues/issues/1354

PR #140, opened 2019-07-11
make sure we got enough random bytes

apreiml:
Fails if not enough random bytes could be read from urandom.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -36,7 +36,11 @@ end
 
 -- Efficient function to get a random string
 function random_string()
-    local random_bytes = io.open("/dev/urandom"):read(64);
+    local length = 64
+    local random_bytes = io.open("/dev/urandom"):read(length);
+    if string.len(random_bytes) ~= length then
+        error("Not enough random bytes read")
+    end
     return tohex(random_bytes);
 end
 


Psycojoker:
Hello,

First, thanks for your contribution <3

I'm surprise by this issue, could you describe a situation where this happen?

apreiml:
Hi :),

this didn't happen. I wrote this patch mainly because I was reading https://nostarch.com/seriouscrypto and there was a section about what can go wrong while getting random numbers. It cited a source from libressl, where retried getting random numbers until enough bytes where gathered: https://books.google.at/books?id=hLcrDwAAQBAJ&pg=PT84&lpg=PT84&dq=random_bytes_safer&source=bl&ots=PZmVJOwlQk&sig=ACfU3U1VVF8tn3MCBzMHxWd3iyVAQwHlvw&hl=en&sa=X&ved=2ahUKEwiSiLbt2q7jAhWIposKHVCkDxQQ6AEwAHoECAQQAQ#v=onepage&q=random_bytes_safer&f=false

The lua read call is successful if it read less bytes than specified. It may not be likely to go wrong, but if it really goes wrong it would be better to fail at least.

Psycojoker:
Ah, need to dig into that (or if someone else wants to do that) but that looks good '-'

Thx a lot for this patch :heart:

Psycojoker:
Finally had the time to test that (in a shell and in sso), seems to be working at expected, thx again! :+1:

PR #141, opened 2019-07-14
Fixed: bug serving files because of LUA exception

moutonjr:
"Interrupted system call" during PWDIR change on "popen" function
Restored by removing FIND's wildcard which could yield fo emptiness
(example : `mydir/` is empty; making `cd mydir && find *` yielding
to non argumented FIND function.

Tested on Stretch with nginx (1.10.3-1+deb9u2) and lua5.1 (5.1.5-8.1+b2).

Note that this could lead to bugs, code refactor should enclose lua-filesystem (1.6.3-1),
but I'm a rookie in LUA.

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -317,10 +317,8 @@ end
 --
 
 function scandir(directory, callback)
-    -- FIXME : this sometime fails (randomly...)
-    -- because of 'interrupted system call'
     -- use find (and not ls) to list only files recursively and with their full path relative to the asked directory
-    local pfile = io.popen('cd "'..directory..'" && find * -type f')
+    local pfile = io.popen('find "'..directory..'" -type f')
     for filename in pfile:lines() do
         callback(filename)
     end


moutonjr:
(note that this crashed some of my apps behind SSOWat, e.g. rainloop, yeilding to HTTP/500 on loadings)

decentral1se (approved):
Looks legit then if I understood correctly!

```
➜  ~ mkdir test 
➜  ~ cd test && find * -type f
zsh: no matches found: *
[1]    8291 exit 1     find * -type f
➜  test echo $?
1
➜  test cd ..                    
➜  ~ find test -type f        
➜  ~ echo $?
0
```

alexAubin:
Oh I didn't saw that PR earlier

That's an interesting fix, thank you :astonished: !

alexAubin (approved):
Not tested but LGTM

PR #78, opened 2017-05-13
Logging by sidddy

Psycojoker:
Hello,

As said there https://github.com/sidddy/SSOwat/commit/e6b220dce0b3de011995662c51c68875526507ef#commitcomment-22128490 the logging work by @sidddy looks very interesting. To avoid losing this work I'm creating this PR based on this commit (+ removing unrelated things) to avoid losing track of it.

I haven't tested it.

PR #145, opened 2019-10-03
Logging reloaded

alexAubin:
Working on finishing https://github.com/YunoHost/yunohost/pull/797 where we need to rework part of the logic of protected/unprotected url ... Thought I would need some sort of straightforward mechanism to get logs from inside SSOwat's Lua code, so ended up resurrecting https://github.com/YunoHost/SSOwat/pull/78 ...

So with this PR, SSOwat logs debug messages in `/var/log/nginx/ssowat.log`, though nothing will be logged by default and if you actually want to enable it you need to add `logging = "debug"` to ssowat's configuration. (We probably don't want to enable all that logging by default because that's way too much information and there's also metadata / privacy consideration)

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -8,6 +8,10 @@
 -- another.
 --
 
+-- Path of the configuration
+conf_path = "/etc/ssowat/conf.json"
+log_file = "/var/log/nginx/ssowat.log"
+
 -- Remove prepending '@' & trailing 'init.lua'
 script_path = string.sub(debug.getinfo(1).source, 2, -9)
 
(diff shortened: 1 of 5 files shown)


init.lua
```diff
@@ -54,8 +63,5 @@ for file in lfs.dir(locale_dir) do
     end
 end 
 
--- Path of the configuration
-conf_path = "/etc/ssowat/conf.json"
-
 -- You should see that in your Nginx error logs by default
 ngx.log(ngx.INFO, "SSOwat ready")
```
kay0u, line 67: Why you don't use ``` local logger = require("log")``` here? (noob here)

alexAubin: Not sure either, I tried at some point, but this init.lua is kinda special because that's where the conf_path gets defined, which is needed to load the `log` module ... and also there's the thing that if you want people to be able to tweak ssowat settings without having to restart/reload nginx, you need to load the settings outside of `init.lua` ... (Not 100% clear for me, sorry)

PR #147, opened 2019-10-03
More extensive check between allowed rules vs. protected rules

alexAubin:
This is a follow-up of [this discussion](https://github.com/YunoHost/yunohost/pull/797#issuecomment-533684426) to try to fix this thing where if you have these settings : 

```
    "protected_urls": [
        "yolo.test/foo", 
        "yolo.test/foo/admin"
    ], 
    "users": {
        "alice": {
            "yolo.test/foo": "Foo Test", 
            "yolo.test/foo/admin": "Foo Test (Admin)"
        }, 
        "bob": {
            "yolo.test/foo": "Foo Test"
        }
    }
```

then currently `bob` has access to `/foo/admin` despite the fact that the intention would be that only `alice` should ...

This draft allow to perform a more extensive check, and by this I mean checking what's the rule that has the longest match with respect to the requested uri...

This seem to be kinda working but gotta run moar test to validate the thousand of different use case to see that this does not cause any regression

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -208,17 +208,15 @@ end
 -- If the URL matches one of the `redirected_urls` in the configuration file,
 -- just redirect to the target URL/URI
 --
--- A match function that uses PCRE regex as default
+-- The 'match' function uses PCRE regex as default
 -- If '%.' is found in the regex, we assume it's a LUA regex (legacy code)
+-- 'match' returns the matched text.
 function match(s, regex)
     if not string.find(regex, '%%%.') then
-        if rex.match(s, regex) then
-            return true
-        end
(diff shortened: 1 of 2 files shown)


alexAubin:
So that's kinda working, still would be nice if anybody has feedback if there's a better way to implement this ...

Josue-T (approved):
For me it look like good. But I didn't a really deep review...

I still just not really like the idea to merge the full url and the "url without the domain".

Maybe there are a better way to do this (maybe as discussed [here](https://github.com/YunoHost/yunohost/pull/797#issuecomment-533748055) ) but we still can improve this later. I think it's just some internal change. So it won't impact the apps.

PR #148, opened 2019-10-27
[fix] CVE-2019-11043

zamentur:
I suggest this to fix CVE-2019-11043 .

Note: it's a theorical fix, i think it should fix the exloit but for instance i hadn't tested it...

I have just tested that

```
curl -i -X GET --url http:/DOMAIN/%0d
curl -i -X GET --url http://DOMAIN/%0a
```
return FORBIDEN

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -30,7 +30,6 @@ ngx.header["X-SSO-WAT"] = "You've just been SSOed"
 
 -- Quick fix for CVE-2019-11043 about php-fpm and nginx
 local loweruri = string.lower(ngx.var.uri)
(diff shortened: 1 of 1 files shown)


zamentur:
Some explanation about this bug https://meterpreter.org/cve-2019-11043-php-fpm-arbitrary-code-execution-vulnerability-alert/

alexAubin:
Alright folks idk we just ended up advising people to upgrade their PHP packages shipped by Debian and it's pretty touchy to touch this part of SSOwat's code ..

Feel free to reopen if you think that's relevant nevertheless ..

PR #151, opened 2020-01-17
[FIX] string.match

kay0u:
`runtime error: /usr/share/ssowat/helpers.lua:330: attempt to call global 'match' (a nil value)`

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -22,9 +22,6 @@ local conf = config.get_config()
 -- Import helpers
 local hlp = require "helpers"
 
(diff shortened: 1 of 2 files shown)


Psycojoker:
I'm surprised this error got undetected for so long, maybe we can try to find if some lua linter could check for those kind of errors on our code in case there are more?

kay0u:
The error occurs only when `protected_regex` is defined, not in so much apps.

Psycojoker:
Indeed.

Thanks for that :)

PR #153, opened 2020-01-20
Fix legacy/new permissions

### LINKED ISSUE #1507: Make (un)protected_url/regex works again
TODO:
- [x] Fix protected_uris / protected_regex not working anymore
- [ ] User with no main permission are not able to answer to link send that are unprotected

Here are old and new ssowat conf file.

stable:
> yunohost: 
  repo: stable
  version: 3.6.5.3
yunohost-admin: 
  repo: stable
  version: 3.6.5.1
moulinette: 
  repo: stable
  version: 3.6.4.1
ssowat: 
  repo: stable
  version: 3.6.4

[/etc/ssowat/conf.json](https://paste.yunohost.org/tuqaquyoxa.json)
```json
{
[...]
    "protected_regex": [
        "stable%.tld/bozon/index%.php$", 
        "stable%.tld/bozon/index%.php%?p=.*$"
    ], 
    "protected_urls": [
        "stable.tld/dotclear2/admin"
    ], 
[...]
    "unprotected_regex": [], 
    "unprotected_urls": [
        "stable.tld/bozon", 
        "stable.tld/dotclear2"
    ], 
    "users": {
        "kayou": {
            "stable.tld/bozon": "BoZoN", 
            "stable.tld/dotclear2": "DotClear 2"
        }
    }
}
```

unstable:
> yunohost: 
  repo: local
  version: 3.7.0.4+202001031317
yunohost-admin: 
  repo: local
  version: 3.7.0.1+202001031053
moulinette: 
  repo: local
  version: 3.7.0+202001031317
ssowat: 
  repo: local
  version: 3.7.0.1+202001031317

[/etc/ssowat/conf.json](https://paste.yunohost.org/gowokafahi.json)
```json
{
[...]
    "permissions": {
        "/": [
            "kayou"
        ]
    }, 
[...]
    "protected_regex": [
        "domain%.tld/bozon/index%.php$", 
        "domain%.tld/bozon/index%.php%?p=.*$"
    ], 
    "protected_urls": [
        "domain.tld/dotclear2/admin"
    ],
[...]
    "unprotected_regex": [], 
    "unprotected_urls": [
        "domain.tld/bozon", 
        "domain.tld/dotclear2"
    ], 
    "users": {
        "kayou": {
            "domain.tld/bozon": "BoZoN", 
            "domain.tld/dotclear2": "DotClear 2"
        }
    }
}
```

In both case I have `bozon` and `dotclear2` installed.

For bozon:
- 3.6 `protected_regex` makes index.php unreachable for visitors, but kayou can access it.
- 3.7 we can't access the app since no permission allow the user kayou to access to the `protected_regex` `domain%.tld/bozon/index%.php$`

For dotclear2:
- The root is accessible for visitors and kayou
- 3.6 the admin part is accessible only by kayou
- 3.7 the admin part is not accessible by kayou since no permission allow him to access to the part.

To avoid regression between 3.6 and 3.7, we should access bozon and admin part of dotclear2 if we are logged in. A logged in user should not have access to a path only if the related permission exists and if he's not in.

For the conf file, imo, the permission entry should have the absolute path:
```json
"permissions": {
        "/": [
            "kayou"
        ]
    }
```
should be:

```json
"permissions": {
        "domain.tld/app_path/": [
            "kayou"
        ]
    }
```

Well I need your help to clarify the situation and to help me to smash this bug!

kay0u:
Needs this [PR](https://github.com/YunoHost/yunohost/pull/871)

Close: https://github.com/YunoHost/SSOwat/pull/152
Resolve: https://github.com/YunoHost/issues/issues/1507
And should resolve all our problems with legacy/new permissions system

I've tested it myself, but someone familiar with this project should indeed do a double triple check

To test it, apply https://github.com/YunoHost/yunohost/pull/871 and this PR, reload nginx (`systemctl reload nginx`) and run `yunohost app ssowatconf`
Finally, play with unprotected/protected/skipped_url

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -334,7 +334,9 @@ function serveThemeFile(filename)
   serveAsset("/ynhtheme/"..filename, "themes/"..conf.theme.."/"..filename)
 end
 
-if hlp.is_logged_in() then
+function serveYnhpanel()
+    logger.debug("Serving ynhpanel")
+
     -- serve ynhpanel files
     serveAsset("/ynh_portal.js", "js/ynh_portal.js")
     serveAsset("/ynh_overlay.css", "css/ynh_overlay.css")
@@ -343,19 +345,7 @@ if hlp.is_logged_in() then
     -- but I didn't succeed to figure out where is the current location of the script
     -- if you call it from "portal/assets/themes/" the ls fails
     scandir("/usr/share/ssowat/portal/assets/themes/"..conf.theme, serveThemeFile)
-
-    -- If user has no access to this URL, redirect him to the portal
-    if not hlp.has_access() then
-        return hlp.redirect(conf.portal_url)
-    end
-
-    -- If the user is authenticated and has access to the URL, set the headers
-    -- and let it be
-    hlp.set_headers()
-    return hlp.pass()
 end
-
-
 
 --
 -- 7. Unprotected URLs
@@ -375,6 +365,8 @@ if conf["unprotected_urls"] then
         or  hlp.string.starts(ngx.var.uri..hlp.uri_args_string(), url))
         and not is_protected() then
             if hlp.is_logged_in() then
+                serveYnhpanel()
+
                 hlp.set_headers()
             end
             logger.debug(ngx.var.uri.." is in unprotected_urls")
@@ -389,6 +381,8 @@ if conf["unprotected_regex"] then
         or  hlp.match(ngx.var.uri..hlp.uri_args_string(), regex))
         and not is_protected() then
             if hlp.is_logged_in() then
+                serveYnhpanel()
+
                 hlp.set_headers()
             end
             logger.debug(ngx.var.uri.." is in unprotected_regex")
@@ -397,6 +391,20 @@ if conf["unprotected_regex"] then
     end
 end
 
+
+if hlp.is_logged_in() then
+    serveYnhpanel()
+
+    -- If user has no access to this URL, redirect him to the portal
+    if not hlp.has_access() then
+        return hlp.redirect(conf.portal_url)
+    end
+
+    -- If the user is authenticated and has access to the URL, set the headers
+    -- and let it be
+    hlp.set_headers()
+    return hlp.pass()
+end
 
 
 --
@@ -452,6 +460,6 @@ end
 -- when trying to access http://main.domain.tld/ (SSOwat finds that user aint
 -- logged in, therefore redirects to SSO, which redirects to the back_url, which
 -- redirect to SSO, ..)
-logger.debug("No rule found for this url. By default, redirecting to portal")
+logger.debug("No rule found for "..ngx.var.uri..". By default, redirecting to portal")
 local back_url = "https://" .. ngx.var.host .. ngx.var.uri .. hlp.uri_args_string()
 return hlp.redirect(conf.portal_url.."?r="..ngx.encode_base64(back_url))
--- a/helpers.lua
+++ b/helpers.lua
@@ -267,95 +267,54 @@ end
 function has_access(user)
     user = user or authUser
 
-    if not conf["users"][user] then
-        conf = config.get_config()
-    end
+    -- Get the longest url permission
+    longest_permission_match = longest_url_path(permission_matches()) or ""
 
-    -- If there are no `users` directive, or if the user has no ACL set, he can
-    -- access the URL by default
-    if not conf["users"] or not conf["users"][user] then
-        logger.debug("No access rules defined for user "..user..", assuming it can access..")
+    logger.debug("Longest permission match : "..longest_permission_match)
+
+    -- If no permission matches, it means that there is no
+    -- permission defined for this url, a logged-in user can access it.
+    if longest_permission_match == "" then
+        logger.debug("No access rules defined for user "..user..", assuming it can access.")
         return true
     end
 
-    -- Loop through user's ACLs and return if the URL is authorized.
-    allowed_url_matches = {}
-    for url, app in pairs(conf["users"][user]) do
+    -- All user in this permission
+    allowed_users = conf["permissions"][longest_permission_match]
 
-        -- Replace the original domain by a local one if you are connected from
-        -- a non-global domain name.
-        if ngx.var.host == conf["local_portal_domain"] then
-            url = string.gsub(url, conf["original_portal_domain"], conf["local_portal_domain"])
-        end
-
-        if string.ends(url, "/") then
-            url = string.sub(url, 1, -1)
-        end
-
-        if string.starts(ngx.var.host..ngx.var.uri, url) then
-            logger.debug("User is allowed to access this match : "..url)
-            table.insert(allowed_url_matches,url)
+    -- The user has permission to access the content if he is in the list of this one
+    if allowed_users then
+        for _, u in pairs(allowed_users) do
+            if u == user then
+                logger.debug("User "..user.." can access "..ngx.var.uri)
+                log_access(user, longest_permission_match)
+                return true
+            end
         end
     end
 
-    -- Keep only the longest match and compare it to the longest protected
-    -- match e.g. we don't want to allow the user to access /foo/admin if
-    -- /foo/admin is protected, but this user is only allowed to access /foo
-    local longest_allowed_match = longest_url_path(allowed_url_matches) or ""
-    local longest_protected_match = longest_url_path(protected_matches()) or ""
-
-    logger.debug("Longest allowed match : "..longest_allowed_match)
-    logger.debug("Longest protected match : "..longest_protected_match)
-
-    -- For the user to be able to access the content, at least one rule should
-    -- exist and it should be the longest match
-    if longest_allowed_match ~= ""
-    and string.len(longest_allowed_match) >= string.len(longest_protected_match) then
-        logger.debug("Logged-in user can access "..ngx.var.uri)
-        log_access(user, longest_allowed_match)
-        return true
-    else
-        logger.debug("Logged-in user cannot access "..ngx.var.uri)
-        return false
-    end
+    logger.debug("User "..user.." cannot access "..ngx.var.uri)
+    return false
 end
 
-
-function protected_matches()
-    if not conf["protected_urls"] then
-        conf["protected_urls"] = {}
-    end
-    if not conf["protected_regex"] then
-        conf["protected_regex"] = {}
+function permission_matches()
+    if not conf["permissions"] then
+        conf["permissions"] = {}
     end
 
     local url_matches = {}
 
-    for _, url in ipairs(conf["protected_urls"]) do
+    for url, permission in pairs(conf["permissions"]) do
         if string.starts(ngx.var.host..ngx.var.uri..uri_args_string(), url)
         or string.starts(ngx.var.uri..uri_args_string(), url) then
-            logger.debug("protected_url match current uri : "..url)
+            logger.debug("Url permission match current uri : "..url)
+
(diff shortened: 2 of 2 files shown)


kay0u:
Ping @maniackcrudelis

This PR should fix the backwards compatibility with legacy permission. If you have time to test it.. I just don't want to create a big vulnerability by omitting something

maniackcrudelis:
Well... I just made a simple test about the thing I was worried about. Results aren't great...

I installed Jirafeau and Lutim, both have some specific SSO rules.
See https://github.com/YunoHost-Apps/lutim_ynh/blob/master/scripts/install#L195-L205 and https://github.com/YunoHost-Apps/jirafeau_ynh/blob/testing/scripts/install#L150-L156

As expected, these url are private:
    https://domain.tld/jirafeau/
    https://domain.tld/jirafeau/admin.php
    https://domain.tld/lutim/
    https://domain.tld/lutim/stats/
    https://domain.tld/lutim/manifest.webapp/
These ones are public:
    https://domain.tld/jirafeau/f.php?h=1iK2k-Nl
    https://domain.tld/lutim/fd34obVg/s2KOOupx.jpg
(I uploaded some files for the test)

Then I moved to testing 3.7
All private url for Lutim are now public.
    https://domain.tld/lutim/
    https://domain.tld/lutim/stats/
    https://domain.tld/lutim/manifest.webapp/
Into the sso conf file, a line has been added for `unprotected_urls` as following:
```
"domain.tld/lutim"
```
Removing that new line does not fix the issue.

Then, I applied this fix but it didn't change anything. Neither the new line in sso config or the app now publicly accessible.
So, lutim became publicly accessible without any notice by upgrading to testing.

Also, as I just checked the doc (https://github.com/YunoHost/doc/blob/group-and-permissions/groups_and_permissions.md#migrating-away-from-the-legacy-permission-management) except saying that it won't work anymore, I have no clues about how to fix that.
And, as explained on the forum (If I remember well) to move from public to private an user should now play with 5 different permissions if a script is not doing it for him. Which is clearly not acceptable...

zamentur:
So if i understand well the issue with lutim:



The migration add "lutim main permission" to visitors. It's seem to be translated as "unprotected" rules (not 100% sure).

And the bug would be protected_regex are not considered, may be because unprotected on / is defined ??? I remember discussions about ordering permissions url rules with length of url prefix... But in the regex case it's not possible.

I am reading the ssowat code to try to understand

maniackcrudelis:
I did remove the `unprotected_urls` added for lutim but it didn't solve anything.
It's not removed in the conf.json I sent you though

zamentur:
i think you have not regenerate the ssowat conf with the fix https://github.com/YunoHost/yunohost/pull/871
or this fix desn't work ?
because the url in "permission" is still "/" in your ssowat conf: https://paste.yunohost.org/loletucuji

maniackcrudelis:
That ssowat conf https://paste.yunohost.org/loletucuji is without the fix. Just the conf after the upgrade to testing.
That other one https://paste.yunohost.org/apikomuhiy is after applying the fix.

zamentur:
Even if it worked i think use skipped_uri on / and protected_uri or regex on a subpath is very strange.

skipped_uris means "SSOWat don't manage this path.". So subpath should be concerned too...

zamentur:
Found it! Indeed, in stable we do "3 redirect, 4 protected, 5 skipped" and in unstable we do "3 redirect, 4 skipped, 5 protected". That's the reason why you have this issue with lutim.

It's a regression, but as explained, it's abnormal to consider to examine protected uris before skipped...

maniackcrudelis:
You mean instead of an `unprotected_urls` ?
I don't remember exactly why as it's an old part of the code. But I think it was because we had an issue with some part of the code around here in the sso.
And since it works that way, I guess I never changed it back.

maniackcrudelis:
Using `unprotected_uris` does work indeed before and after the upgrade.

maniackcrudelis:
By the way, trying the permission from the new panel, I notice that I can remove the visitors permission, which make the images private.
But there's no way to make the app public, since that the protected_regex that make the app private.

Anyway, from this fix it seems to work after an upgrade. Even if we noticed a possible regression.
But still, this feature needs to be more clear about some other points, like the doc, a proper way to migrate and the usage of script to move from public to private.

So as I asked on the chat, a meeting would be welcome to address those questions.

zamentur (changes requested):
This PR seems to kill all the backward compatibility. For example an old app package using protected_uris instruction install with a 3.7 and this change will not do the correct thing (i think).

Alternatively, i suggest to fix https://github.com/YunoHost/issues/issues/1507 by modifying the migration and the app_addaccess.

kay0u:
> Found it! Indeed, in stable we do "3 redirect, 4 protected, 5 skipped" and in unstable we do "3 redirect, 4 skipped, 5 protected". That's the reason why you have this issue with lutim.

I'm not sure about that:
Stable: https://github.com/YunoHost/SSOwat/blob/stretch-stable/access.lua#L256

Unstable: https://github.com/YunoHost/SSOwat/blob/stretch-unstable/access.lua#L281
in both case, it's a function, and this function is called in the `Unprotected URLs`, during the point 7

kay0u:
I think I found the "bug":

In the [stable version](https://github.com/YunoHost/SSOwat/blob/0c55f38fda529c02ae3f350142619a09d646b593/access.lua#L293), `skipped_urls/regex` check if this path is `protected`... It looks super weird since the skip should be used to... skip --'
In the [unstable version](https://github.com/YunoHost/SSOwat/blob/12412cc6c89083b185d627e2832539e54adec25e/access.lua#L254), we don't check if this path is protected.

If I re add the check in the skip url, it fix the "bug", but I'm not sure if it's a bug or not...

kay0u:
I tried to change `skipped_uris` into `unprotected_uris` in lutim scripts, and it looks good to me. (don't forget to remove `skipped_uris from the settings.yml of this app and run `yunohost app ssowatconf` if you want to try on your side)

kay0u:
> By the way, trying the permission from the new panel, I notice that I can remove the visitors permission, which make the images private.
> But there's no way to make the app public, since that the protected_regex that make the app private.
> 
> Anyway, from this fix it seems to work after an upgrade. Even if we noticed a possible regression.
> But still, this feature needs to be more clear about some other points, like the doc, a proper way to migrate and the usage of script to move from public to private.
> 
> So as I asked on the chat, a meeting would be welcome to address those questions.

The current limitation of permissions is indeed regex (we just can't for now). It can evolve ofc, and we can implement it if we need it.

A meeting would be great to fix what we want from permission system.

The documentation will come after the permissions system is fully implemented and when everyone is in agreement with how it works.

zamentur:
> I think I found the "bug":

"It's not a bug, it's a feature"
https://github.com/YunoHost/SSOwat/commit/247847a20354a4ada50e57f8a71fe31a4ed013d3

So the question is do we want this feature now. I think yes.
I add this topic  for the meeting https://pad.lqdn.fr/p/yunohost-28-01-2020

zamentur (changes requested):
To finish this work i think we should move this condityion https://github.com/YunoHost/SSOwat/blob/stretch-unstable/access.lua#L348 after the unprotected section.

Indeed if a user has no access to an app because his name is not in the list of allowed user in a permission, he should be able for example to answer to poll send by other users like a visitor could do (unprotected uris/regex).

I have read the other change and i am finally agree with its.

kay0u:
> Indeed if a user has no access to an app because his name is not in the list of allowed user in a permission, he should be able for example to answer to poll send by other users like a visitor could do (unprotected uris/regex).

You're right! I'll change that

> I have read the other change and i am finally agree with its.

In fact, I found a bug with [these lines](https://github.com/kay0u/SSOwat/blob/fix-legacy-permissions/helpers.lua#L277-L279).
When I try to go on the sso, I browse https://domain.tld/, in my PR there is no permission for that, so the user is by default allowed . And, I don't reach the sso but the default nginx index.html.
Same, when I try to access an url that doesn't exist, I reach the default nginx index.html, not the sso.

kay0u:
Bump

Josue-T:
Hello, sorry for the delay, I'll try to review all PR (on the group/permission) theses next days.

Josue-T (approved):
Look like OK for me.

Have you tested this new code with the yunohost unit test ?

kay0u:
yup, but we don't test any edge cases during unit tests.

Josue-T:
Hello,

After thinking about the `skipped_urls/regex` and `protected` priority I was thinking about that and for me:
- the difference between `skipped_urls` and `protected_urls` is just about the user authentication management (In the HEADER). So for me there are no reason to drop the possiblitly to protect a sub-url of a `skipped_urls`.
- with this new change we are less consistant. Why we have different priority of the permission if we change the `skipped_urls` by `unprotected_urls`.

By example I'll working to implement a new SSO management for synapse. Actually we have `/` set as `skipped_urls` because it's just a server which provide a server API and client API, so I really don't want set this as unprotected. When I'll implement the SSO I will add small php code which will authenticate the user from the HEADER `REMOTE_USER`. To have this to work I'll need to set this as `protected`. So we will have this configuration by example:
`domain.tld/_matrix/` -> `skipped_urls`.
`domain.tld/_matrix/my_sso/auth.php` -> `protected`.

So for all of theses reason I think that it's better to keep the old way to mange this.

kay0u:
I don't have any strong opinion on it, I can add it again if we feel the need.

kay0u:
The reason why this change have been done: https://github.com/YunoHost/SSOwat/pull/149

~So, I don't know what to do now >.<~

I propose to fix this bug on another PR, since I didn't change this part, and as I would like to see this one merged (if it works as intended)

kay0u:
This PR should fix most of remains issue with legacy permissions:
https://github.com/kay0u/SSOwat/commit/af892991afda2f8cd9449606bf8600838a931db8

(@Josue-T use it here: https://github.com/YunoHost-Apps/synapse_ynh/pull/175)

I use a lot of that PR, but it's not completely related to this one. Should I merge it here or create a new one that contains the fix and this PR?

kay0u:
This PR should fix most of remains issue with legacy permissions:
https://github.com/kay0u/SSOwat/commit/af892991afda2f8cd9449606bf8600838a931db8

(@Josue-T use it here: https://github.com/YunoHost-Apps/synapse_ynh/pull/175)

I use a lot of that PR, but it's not completely related to this one. Should I merge it here or create a new one that contains the fix and this PR?

kay0u:
This PR should fix most of remains issue with legacy permissions:
https://github.com/kay0u/SSOwat/commit/af892991afda2f8cd9449606bf8600838a931db8

(@Josue-T use it here: https://github.com/YunoHost-Apps/synapse_ynh/pull/175)

I use a lot of that PR, but it's not completely related to this one. Should I merge it here or create a new one that contains the fix and this PR?

kay0u:
This PR should fix most of remains issue with legacy permissions:
https://github.com/kay0u/SSOwat/commit/af892991afda2f8cd9449606bf8600838a931db8

(@Josue-T use it here: https://github.com/YunoHost-Apps/synapse_ynh/pull/175)

I use a lot of that PR, but it's not completely related to this one. Should I merge it here or create a new one that contains the fix and this PR?

Josue-T:
> Should I merge it here or create a new one that contains the fix and this PR?

For me it's equals. I would say that it's better to create an other PR, to make more easy to review even if it's based on this one.

kay0u:
Closing this issue, replaced by https://github.com/YunoHost/SSOwat/pull/156

PR #154, opened 2020-02-21
Fix auth error for some apps

yalh76:
Fixing https://github.com/YunoHost/issues/issues/1420

For now, need to manually add in /etc/ssowat/conf.json:
```
    "noauth_urls": [
        "url with auth disabled"
    ], 
```


yalh76:
switching to https://github.com/YunoHost/SSOwat/pull/155

PR #155, opened 2020-02-22
Noauth

### LINKED ISSUE #1420: Add setting for apps to disable Basic HTTP Auth header injection in some uris
This is the cause of the issue of "Invalid Token" that is regularly reported by users of the Mastodon_ynh, Peertube_ynh, Bitwarden_ynh, and other apps.

SSOwat injects some basic HTTP auth header here : https://github.com/YunoHost/SSOwat/blob/stretch-unstable/helpers.lua#L394-L397

This conflicts with the oauth thing in peertube (dunno about other apps, we only investigated peertube with @yalh76) which will check if there's an Authorization header in the request, and if there is, assume it's a Bearer / Token auth header and won't try to check if it's a basic HTTP auth instead : 

https://github.com/oauthjs/node-oauth2-server/blob/master/lib/handlers/authenticate-handler.js#L111-L122

https://github.com/oauthjs/node-oauth2-server/blob/master/lib/handlers/authenticate-handler.js#L141-L147

So it's not clear if it's a bug in SSOwat or OAuth module ...

But anyway, I think a simple fix would be to add an option similar to `unprotected_uri` that could be called `dont_inject_auth_header_uri` to disable the Authorization header injection for these.

That of course raises the bigger question of how we're supposed to support SSO support for these app since it really is dependent on Basic HTTP auth support in the upstreams...

yalh76:
Fix https://github.com/YunoHost/issues/issues/1420

need to be used with https://github.com/YunoHost/yunohost/pull/883

The issue addressed is:
- when the app is private
- when the app is public and you are logged in YunoHost

Maybe an array would help:
Usual way:
|         | Not Logged             | Logged              |
|---------|------------------------|---------------------|
| Public  | Content without header | Content with header |
| Private | Go to portal           | Content with header |

Using noauth: managing content without header when logged:
|         | Not Logged             | Logged                 |
|---------|------------------------|------------------------|
| Public  | Content without header | Content **without** header |
| Private | Go to portal           | Content **without** header |


access.lua
```diff
@@ -262,6 +264,23 @@ and string.len(longest_skipped_match) > string.len(longest_unprotected_match) th
...
+-- it means that the URL should be protected by the SSO but no header
+-- has to be sent, even if the user is already authenticated.
+--
+
+if longest_noauth_match ~= ""
+and string.len(longest_noauth_match) >= string.len(longest_protected_match) 
+and string.len(longest_noauth_match) > string.len(longest_unprotected_match) then
+    if hlp.is_logged_in() then
+      logger.debug("Noauth "..ngx.var.uri)
+      return hlp.pass()
```
alexAubin, line 280: I guess technically speaking this mechanism "works". But overall it seems like you're just duplicating the `skipped_uri` mechanics with a different name ? So my understanding is that that doesn't really address the original issue, which is about being unable to have the app working as expected if using `unprotected_uri` (or `protected_uri`, i.e. the app as 'private' with the new mechanism) because it will just be as if the app was skipped.

Instead, I believe this new `noauth_uri` thing should only be applied [here](https://github.com/YunoHost/SSOwat/blob/stretch-unstable/helpers.lua#L502) with something like : 

```
if [url is not in noauth_uris]
then
    -- Set the authorization header, except if the uri is in the noauth_uris
    ngx.req.set_header("Authorization", "Basic " [...etc...])
end
```

That way, the whole `noauth` thing remains independent from the whole protected/unprotected/(skipped) thing.

(Of course actually implementing it in lua is another thing as I'm not so much at ease with that language @.@)

alexAubin (changes requested):

access.lua
yalh76: Yes the issue addressed is:
- when the app is private
- when the app is public and you are logged in YunoHost

Well its' a bit different from `skipped_uri`, with `skipped_uri` an app is always public, the content of the page is sent with `return hlp.pass()` without headers : https://github.com/YunoHost/SSOwat/blob/4f692c99d81164848975b32a501eaf688147c153/access.lua#L258-L263

In the case of noauth, content is only sent without headers or anything else when the user is logged in: L278 `if hlp.is_logged_in() then`

Maybe an array would help:
Usual way:
|         | Not Logged             | Logged              |
|---------|------------------------|---------------------|
| Public  | Content without header | Content with header |
| Private | Go to portal           | Content with header |

Using noauth: managing content without header when logged:
|         | Not Logged             | Logged                 |
|---------|------------------------|------------------------|
| Public  | Content without header | Content **without** header |
| Private | Go to portal           | Content **without** header |

&nbsp;

Your proposition was my first idea but as far I've seen and as I don't know lua, the only way would be to add arguments to the `function set_headers(user)`, like `function set_headers(user,uri,noauth_uris)` and it seems to me a bit to heavy as `set_headers(user)` is used several times in access.lua

alexAubin: Alrighty, that makes more sense than I tought, sorry for not catching that subtlety first time.

Though imho we should still add an `and hlp.has_access()` to that `if` three lines above, as an app (or url) may be accessible only to specific users.

Apart from this, looks good to me :+1:

yalh76: That was very subtil and I didn't provided enough comments to make it understandable ... 
And that PR comes in the middle of permissions implementations.... No very easy ^^
I will look at the `and hlp.has_access()` and let you know

yalh76: `and hlp.has_access()` added

alexAubin (approved):
LGTM ¯\\_\(ツ)\_/¯

yalh76:
As @Josue-T is working to integrate noauth in https://github.com/YunoHost/yunohost/pull/861, I will close that PR :)

PR #156, opened 2020-03-03
Refactor skipped/protected/unprotecter

kay0u:
Replace #153
Related: #149 and YunoHost-Apps/ttrss_ynh/issues/70

The idea is to get the longest protected/unprotected/skipped_url/regex, and to keep only the longest one among them

(the code to review is only the latest commit af89299)

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -239,6 +239,14 @@ if conf["redirected_regex"] then
 end
 
 
+local longest_protected_match = hlp.longest_url_path(hlp.get_matches("protected")) or ""
+local longest_skipped_match = hlp.longest_url_path(hlp.get_matches("skipped")) or ""
+local longest_unprotected_match = hlp.longest_url_path(hlp.get_matches("unprotected")) or ""
+
+logger.debug("longest skipped "..longest_skipped_match)
+logger.debug("longest unprotected "..longest_unprotected_match)
+logger.debug("longest protected "..longest_protected_match)
+
 --
 -- 4. Skipped URLs
 --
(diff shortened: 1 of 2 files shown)


helpers.lua
```diff
@@ -267,96 +267,89 @@ end
...
-        log_access(user, longest_allowed_match)
-        return true
-    else
-        logger.debug("Logged-in user cannot access "..ngx.var.uri)
-        return false
+    local url_matches = {}
+
+    for url, permission in pairs(conf["permissions"]) do
+        if string.starts(ngx.var.host..ngx.var.uri..uri_args_string(), url)
+        or string.starts(ngx.var.uri..uri_args_string(), url) then
```
Josue-T: As I understand here you check here only on the path without the domain. I'm a little bit confused about that. Does it mean that if a packager create a permission with an url `/abcd`. The url will be accessible from all yunohost domain ?

helpers.lua
```diff
@@ -267,96 +267,89 @@ end
 function has_access(user)
```
Josue-T, line 267: About this comment:
```lua
-- Check whether a user is allowed to access a URL using the `users` directive
```
Shouldn't it be mor the `permissions` directive now ?

access.lua
```diff
@@ -247,66 +255,14 @@ end
...
-        or hlp.match(ngx.var.uri..hlp.uri_args_string(), regex) then
-            logger.debug(ngx.var.uri.." is in protected_regex")
-            return true
-        end
-    end
-
-    logger.debug(ngx.var.uri.." is not in protected_urls/regex")
-    return false
+if longest_skipped_match ~= ""
+and string.len(longest_skipped_match) >= string.len(longest_protected_match) then
```
Josue-T: Shouldn't we add an other condition like that:
```lua
string.len(longest_skipped_match) > string.len(longest_protected_match)
```

Josue-T (commented):
Globally it look really good for me. It might be a good improvement.

access.lua
kay0u: Why should we add this condition?

helpers.lua
kay0u: This line is a bad copy/pasta from other rules.

access.lua
Josue-T: Oh my bad, I wrote the bad test that I think...
It was this that I thought:
```lua
string.len(longest_skipped_match) > string.len(longest_unprotected_match)
```

Because we have 3 type of "url match". Here we check if the skipped url is the url which is the best match. To be sure about that we need to compare with the 2 other.

By example we have : 
- `longest_unprotected_match` : `/pathA/api`
- `longest_skipped_match` : `/pathA/`.

In this case the unprotected should be a priority.

kay0u: :100:

Josue-T (approved):

Josue-T:
Hello,

Thinking about the regex support for the permission, I found maybe a solution which will better than the [actual hook](https://github.com/YunoHost/yunohost/pull/877).

Maybe the solution could be to just check if one the permission regex match before all other urls (in line [249](https://github.com/YunoHost/SSOwat/pull/156/files#diff-80c7526357cd62e0d0b6d80c34850105R249)).

kay0u:
I'm not sure I understand exactly what you're proposing... And what it would change on the app side...

kay0u:
I'll merge it in 5 days if no one is against that

PR #159, opened 2020-03-29
Avoid unnecessarily reloading the config file

kay0u:
Fix these tests: https://gitlab.com/yunohost/yunohost/-/jobs/489641336
And a little faster (maybe, I guess?) because we only reload conf one time

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -16,11 +16,12 @@ if not srvkey then
     cache:add("srvkey", srvkey)
 end
 
--- Initialize and get configuration
-local conf = config.get_config()
-
 -- Import helpers
 local hlp = require "helpers"
+
+-- Initialize and get configuration
+hlp.refresh_config()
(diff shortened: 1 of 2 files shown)


alexAubin (approved):
Sounds legit to me, imho we can yolomerge and see if that breaks stuff during alpha/beta-testings

Can you just confirm that it doesn't imply to reload nginx each time ssowat/conf.json (or .persistent) is changed ?

kay0u:
> Sounds legit to me, imho we can yolomerge and see if that breaks stuff during alpha/beta-testings

:heart: 

> Can you just confirm that it doesn't imply to reload nginx each time ssowat/conf.json (or .persistent) is changed ?

Tested with `./ynh-dev test yunohost` :+1:. (The file is read at the beginning of `access.lua` file, not on reload nginx)

PR #160, opened 2020-03-29
closing the files when we're done reading them

kay0u:


Josue-T (approved):

config.lua
```diff
@@ -11,14 +11,16 @@ function get_config()
     -- Load the configuration file
     local conf_file = assert(io.open(conf_path, "r"), "Configuration file is missing")
     local conf = json.decode(conf_file:read("*all"))
-
+    conf_file:close()
```
kay0u: I should check if con_file is not `nil`

alexAubin: eeeh bumpity bump ? apart from this comment this PR looks like easy to merge?

kay0u:
```suggestion
    if conf_file ~= nil then
        conf_file:close()
    end
```

kay0u: Indeed, I think it should works™ ( suggestion not tested)

alexAubin:
Yolomergin after applying Kayou's suggestion

PR #161, opened 2020-03-31
Rework permissions

kay0u:
https://github.com/YunoHost/yunohost/pull/861

~To make it works, I added a `"/",` at the first line of the uris array of `core` permission.
Does it look like a bug? Probably, but at least we can test that way for now.~ Not needed with https://github.com/YunoHost/SSOwat/pull/161/commits/400f88e6cabcb5cac18a39d6373df36084f17ead

Is `re:` before a regex is really necessary?


Josue-T:
> To make it works, I added a "/", at the first line of the uris array of core permission.

I could also add this in the ssowat conf on the core side.

Josue-T:
Hello,

I made some test on this. I've detected that there are a redirection problem when to you type the url `domain.tld/` when you are authenticated.

helpers.lua
```diff
@@ -266,34 +271,60 @@ function log_access(user, uri)
...
+    local longest_url_match = ""
+    
+    for permission_name, permission in pairs(conf["permissions"]) do
+        if next(permission['uris']) ~= nil then
+            for _, url in pairs(permission['uris']) do
+                if string.starts(url, "re:") then
+                    url = string.sub(url, 4, string.len(url))
+                end
+
+                local m = match(ngx.var.host..ngx.var.uri..uri_args_string(), url)
```
Josue-T: I don't know it's a good idea to considere that all url are a regex. Maybe it's better to separate a regex and the a standard url to avoid an url to match because it's managed as a regex.

access.lua
```diff
@@ -304,45 +281,28 @@ function serveYnhpanel()
...
--- It means that you can let anyone access to an app, but if a user has already
--- been authenticated on the portal, he can have his authentication headers
--- passed to the app.
---
+local permission = hlp.get_best_permission()
 
-if longest_unprotected_match ~= ""
-and string.len(longest_unprotected_match) > string.len(longest_protected_match) then
-    if hlp.is_logged_in() then
+if permission then
```
Josue-T: Thinking about this I think that we have some issue here.

The section `7. Basic HTTP Authentication` should be at my point of view before this, because here if the url match a permission not public and there are no user authenticated by a cookie the client is redirected ether if it is authenticated by a header.

I  also think that we need to manage the case if no permission match. Probably we just need to add this `hlp.redirect(conf.portal_url)`. This might avoid the actual loop.

Josue-T (commented):
Good refactoring :+1:

helpers.lua
kay0u: I don't know if it's a problem or not, is this [function](https://github.com/YunoHost/SSOwat/blob/24b3f7dc3a73a5d642e599ea9b671cc90f053270/helpers.lua#L33-L39) fallback to a "normal match" if no regex is found?

access.lua
kay0u: You're probably right, I don't know why that section was there and not before.

I fixed the redirect loop

helpers.lua
Josue-T: > I don't know if it's a problem or not, is this [function](https://github.com/YunoHost/SSOwat/blob/24b3f7dc3a73a5d642e599ea9b671cc90f053270/helpers.lua#L33-L39) fallback to a "normal match" if no regex is found?

I think this function is a little bit different. It's used to give the possibility to support the lua regex and the PCRE regex.

kay0u: Okay, as far as I can see, we're only accepting PCRE regex for permissions ?

Josue-T: The idea is to migrate to PCRE regex but the we still support lua regex for retro-compatibility.

Josue-T:
Hello,

Now, I have [some issue with the SSO cookies in sogo](https://github.com/YunoHost-Apps/sogo_ynh/issues/7#issuecomment-680257907). Do you think that it would be possible to remove all SSO cookies for somes URL ?

kay0u:
You can set the parameter `auth_header` to false for the sogo permission, and it should fix the issue

PR #162, opened 2020-04-16
[fix] invalid more cookies

Psycojoker:
According to roy on xmpp this fix help clearing more cookies (maybe not all) properly.

I don't understand why but I didn't wanted to lose that information.


SilverViper:
Function delete_redirect_cookie also has double ;;'s

Psycojoker:
Superset by https://github.com/YunoHost/SSOwat/pull/163

PR #163, opened 2020-04-30
[fix] Invalid even more cookies

SilverViper:
See: #162

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -177,7 +177,7 @@ function set_auth_cookie(user, domain)
                        "; Expires="..os.date("%a, %d %b %Y %X UTC", expire)..
                        "; Secure"..
                        "; HttpOnly"..
-                       "; SameSite=Lax ;;"
+                       "; SameSite=Lax"
 
     ngx.header["Set-Cookie"] = {
         "SSOwAuthUser="..user..cookie_str,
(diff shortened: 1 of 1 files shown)


helpers.lua
```diff
@@ -214,7 +214,7 @@ function delete_redirect_cookie()
                        "; Expires="..expired_time..
                        "; Secure"..
                        "; HttpOnly"..
-                       "; SameSite=Lax ;;"
+                       "; SameSite=Lax"
     ngx.header["Set-Cookie"] = "SSOwAuthRedirect=;" ..cookie_str
```
kay0u: > Syntax error in Set-Cookie: SSOwAuthRedirect=;; Path=/yunohost/sso/; Expires=Thu, 01 Jan 1970 00:00:00 UTC; Secure; HttpOnly; SameSite=Lax ;; at position 18.

I think we should remove this `;` here (or the one line 213 : `local cookie_str = "; Path="..conf["portal_path"]..`).

kay0u:
What if we left only one `;`?

SilverViper:
> What if we left only one ;?

Response headers do not require trailing ;
[https://developer.mozilla.org/en-US/docs/Glossary/Response_header](url)

alexAubin:
Bumpity bump, anybody has some insight on this ? Shall we yolomerge and see what happens lulz ?

alexAubin:
Sounds like those ;; were initially added by me in https://github.com/YunoHost/SSOwat/commit/2699aa8db727c20dd74fe05887c511fa3f1cf68a but I have no idea why I did this ...

Anyway, yolomerging !

PR #167, opened 2020-06-18
fix theme loading

### LINKED ISSUE #1562: Theming : custom_overlay.css isn't loaded anymore since 3.7
Hi,

On a YNH instance where I have a custom theme, I realized that the YNH button (#ynh-overlay-switch) present in applications was back to the default one.
While inspecting the loaded files, I notice that the custom_overlay.css file is no longer loaded, which explains this.

I tried enabling default themes, I tried on another YNH instance : I have the same behavior. No matter if the app is on a subdomain or subfolder.

yunohost --version on the two instance tested : 
```
yunohost:
  repo: stable
  version: 3.7.1.1
yunohost-admin:
  repo: stable
  version: 3.7.1.1
moulinette:
  repo: stable
  version: 3.7.1
ssowat:
  repo: stable
  version: 3.7.1.1
```

```
yunohost:
  repo: now
  version: 3.7.0.12
yunohost-admin:
  repo: now
  version: 3.7.0.4
moulinette:
  repo: now
  version: 3.7.0.2
ssowat:
  repo: now
  version: 3.7.0.3
```

kay0u:
Fix https://github.com/YunoHost/issues/issues/1562

The command `find /dir -type f` returns a list of absolute path of files.
For example for the theme `clouds`, we get: 
```
$ find /usr/share/ssowat/portal/assets/themes/clouds/ -type f
/usr/share/ssowat/portal/assets/themes/clouds/custom_portal.css
/usr/share/ssowat/portal/assets/themes/clouds/custom_overlay.css
/usr/share/ssowat/portal/assets/themes/clouds/background.jpg
/usr/share/ssowat/portal/assets/themes/clouds/custom_portal.js
/usr/share/ssowat/portal/assets/themes/clouds/cloud.png
```
The [callback](https://github.com/YunoHost/SSOwat/blob/20de3f5f37aa29847c97375cf69f89c188463ce0/access.lua#L277) checks [if the uri matches with `/ynhtheme/the_absolute_path_of_the_file`](https://github.com/YunoHost/SSOwat/blob/20de3f5f37aa29847c97375cf69f89c188463ce0/access.lua#L283), which is never true because we just want to check the filename.

To solve the problem, make sure that "find" only returns filenames.

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -272,7 +272,7 @@ end
 
 function scandir(directory, callback)
     -- use find (and not ls) to list only files recursively and with their full path relative to the asked directory
-    local pfile = io.popen('find "'..directory..'" -type f')
+    local pfile = io.popen('find "'..directory..'" -type f -printf \"%f\n\"')
     for filename in pfile:lines() do
         callback(filename)
     end


alexAubin:
Looks like a small easy fix so yolomerging

PR #169, opened 2020-07-11
Allow the user's password manager to automatically enter the password/username/new-password

manuc66:
Allow the user's password manager to automatically enter the password/username/new-password

See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autocomplete


zamentur (approved):
Thanks for this contribution, LGTM

PR #171, opened 2020-09-07
Update language on each request

kay0u:
Not tested /!\

A quick and dirty fix for this issue: https://forum.yunohost.org/t/cest-du-chinois/12595


alexAubin:
Yolomerging

PR #172, opened 2020-09-20
Moar refactoring (on top of permission rework)

alexAubin:
A bunch of refactoring following review of https://github.com/YunoHost/SSOwat/pull/161

The changes in https://github.com/YunoHost/SSOwat/pull/161 made the code much simpler (way less madness in the handling of unprotected_urls etc...) so I was hoping the whole access.lua code would be easy to read ... but it's still a bit spagetthi because of the whole history of the code

So that's an attempt to make everything even simpler ?

The final structure looks like : 

1. Handle login
2. Handle portal assets
3. Handle redirections (if relevant w.r.t. requested url)
4. Identify most relevant permission to apply
5. Apply permission

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -308,52 +308,40 @@ end
 
 --
 --
+-- 5. APPLY PERMISSION
 --
 --
 
+-- 1st case : client has access
 
+if hlp.has_access(permission) then
 
-if permission then
     if is_logged_in then
-        serveYnhpanel()
+        -- If the user is logged in, we set some additional headers
+        hlp.set_headers()
 
-        -- If the user is authenticated and has access to the URL, set the headers
-        -- and let it be
-        if permission["auth_header"] and hlp.has_access(permission) then
-            logger.debug("Set Headers")
-            hlp.set_headers()
+        -- If Basic Authorization header are disabled for this permission,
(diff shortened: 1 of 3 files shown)


access.lua
```diff
@@ -198,13 +198,40 @@ then
...
+-- 2 ... continued : portal assets that are available on every domains
+--
+-- For example: `https://whatever.org/ynhpanel.js` will serve the
+-- `/yunohost/sso/assets/js/ynhpanel.js` file.
+--
+
+if is_logged_in then
+    assets = {}
+    assets["/ynh_portal.js"] = "js/ynh_portal.js"
+    assets["/ynh_overlay.css"] = "css/ynh_overlay.css"
```
kay0u: According to https://stackoverflow.com/questions/513239/how-to-quickly-initialise-an-associative-table-in-lua
```suggestion
    assets = {
                   ["/ynh_portal.js"] = "js/ynh_portal.js",
                   ["/ynh_overlay.css"] = "css/ynh_overlay.css"
             }
```

kay0u (approved):
I'm not sure that this change is really better... Thanks anyway for the refactoring, the code is close to being... readable \*.\*

PR #176, opened 2020-12-17
feat(security): don't set header if auth_header false

titoko:


alexAubin:
Can you elaborate on the motivation behind that changes ? `set_headers()` actually seem to do a whole bunch of things

https://github.com/YunoHost/SSOwat/blob/dev/helpers.lua#L416-L487

The initial intention of the `auth_header` setting is purely about the `Authorization` header :s

titoko:
Yes, well after talking with Kayou I update my PR :

* Split the set_header method in two method (set_header & refresh_user_cache)

* In the access I don't want to call the set_header method if auth_header is false because some app have their own Header Authorization. In the current situation the Header set by the app is overiddent by the SSO Header & then remove if auth_header is false.

kay0u (approved):
Not tested but LGTM

alexAubin (approved):
Not tested but LGTM on principle :+1:

PR #177, opened 2020-12-23
Change SSOwat auth header to "Proxy-Authorization" to prevent conflict with the app auth header

kay0u:
CoNtExT: I use nextcloud desktop app which use Auth header with a token instead the user password, and ssowat tries to authenticate using this token I receive a [`Connection failed for:`](https://github.com/YunoHost/SSOwat/blob/75cecb9a7aaacf62a47e3e91fba1248ffc6e578b/helpers.lua#L388) in my error.log and I get banned by fail2ban.

[Before version 4.1](https://github.com/YunoHost/SSOwat/blob/ec2d13439faa7d8ea51609f2d0be9607fa576876/access.lua#L345-L373) we parsed the "auth header" after checking that we don't already have access to the url, [in the 4.1](https://github.com/YunoHost/SSOwat/blob/75cecb9a7aaacf62a47e3e91fba1248ffc6e578b/access.lua#L32) we check auth header on each request.

Solution: Move the auth header check at the end.

Half tested, I no longer get banned by fail2ban because there is auth header, but I'm 100% not sure how to test auth header ><

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -263,14 +263,14 @@ function refresh_logged_in()
     return false
 end
 
--- If client set the `Authorization` header before reaching the SSO,
+-- If client set the `Proxy-Authorization` header before reaching the SSO,
 -- we want to match user and password against the user database.
 --
 -- It allows to bypass the cookie-based procedure with a per-request
 -- authentication. This is useful to authenticate on the SSO during
 -- curl requests for example.
 function parse_auth_header()
-    local auth_header = ngx.req.get_headers()["Authorization"]
+    local auth_header = ngx.req.get_headers()["Proxy-Authorization"]
 
     if auth_header then
         _, _, b64_cred = string.find(auth_header, "^Basic%s+(.+)$")
@@ -295,6 +295,9 @@ function parse_auth_header()
                 else
                     return redirect(conf.portal_url)
                 end
+            else
+                -- https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/407
+                ngx.status = 407
             end
         end
     end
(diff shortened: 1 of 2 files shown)


helpers.lua
```diff
@@ -260,35 +260,44 @@ function refresh_logged_in()
...
+                    -- If Basic Authorization header are enable for this permission,
+                    -- add it to the response
+                    if permission["auth_header"] then
+                        set_headers(user)
+                    end
+
+                    return pass()
+                else
+                    return redirect(conf.portal_url)
+                end
```
alexAubin: I guess it works ... but ideally it would be nice if we can keep this part of the logic in access.lua for the sake of clarity :/ Though I'm trying to see how we can do this and it feels like there's no easy / elegant way ...

alexAubin (commented):
I was thinking that part was using the Authorization header :  https://github.com/labriqueinternet/vpnclient_ynh/blob/testing/conf/ynh-vpnclient-loadcubefile.sh#L89

... but in fact nope, sounds like a regular POST + save the cookie ?

So now I'm wondering if that stuff is actually useful / used by anybody ...

helpers.lua
kay0u: We can copy pasta the content for the function parse_auth_header in place we call it in access.lua? (As it was in versions < 4.1

alexAubin:
c.f. the discussion on the chat and tip from leo : I believe we should just rename the header we check into Proxy-Authorization

c.f. https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication

and that should be good ... (and anyway i'm not convinced anybody uses this and it's still possible to auth with curl without this anyway)

JimboJoe:
Tried it and looks good for fixing the initial symptoms! 👍 
Unfortunately, now the portal gives 500 errors, with this message in the main domain nginx errors log file:
```
2020/12/23 23:21:22 [error] 10242#10242: *15550 lua entry thread aborted: runtime error: /usr/share/ssowat/helpers.lua:1063: attempt to concatenate global 'au
thUser' (a nil value)
stack traceback:
coroutine 0:
        /usr/share/ssowat/helpers.lua: in function </usr/share/ssowat/helpers.lua:1055>, client: 192.168.1.254, server: my.domain, request: "GET /yunohost/sso
/?action=logout HTTP/2.0", host: "my.domain", referrer: "https://my.domain/yunohost/sso/?r=aHR0cHM6Ly9tb25pbi5uZXQv"
```

kay0u:
Thank you for the feedback, my latest commit should fix it

alexAubin:
Let's yolomerge and see what happenz /o\

PR #181, opened 2021-03-01
POC: be able to bypass permissions using an admin token

alexAubin:
Soooo today I got fed up of seeing again stuff about having to "unprotect" an app to be able to run some curl requests, then having to protect-it-again-but-we-dont-really-know-if-it-was-protected-or-not-in-the-first-place

That turns a bunch of things that should have been simple into a complex mess.

So instead I'm investigating the idea of : 
- being able to define an admin token to bypass permissions restrictions using a header, let's call it `SSOwat-Admin-Token`
- in `ynh_local_curl`, we temporarily add such a token and inject the corresponding header in the request
- no need to tweak permission before/after ynh_local_curl anymoar
- ???
- PROFIT!

### To test this PR: 

- Pull the corresponding branch, restart nginx
- `install -m 400 -o www-data <(echo secret_token) /etc/ssowat/admin_token`
- install an app in private mode. In my case I installed hextris
- try `curl`ing the page without any auth, this should show the ssowat portal title: `curl -s -L https://yolo.test/hextris --insecure | grep title`
- try adding `--header "SSOwat-Admin-Token: secret_token"`

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -327,7 +327,9 @@ function is_admin()
 
   local token = token:gsub("\n","")
 
-  -- FIXME FIXME FIXME - this is a really stupid and time-attack-prone way to validate the token
+  -- LUA comparison are made in constant time thanks to interned string mechanism
+  -- It compare pointers and not char by char. SO no risk of timing attack here :)
+  -- See https://poprocks.dev/constant-time-string-comparison-in-lua/
   if admin_token_header == token then
       return true
   else
--- a/helpers.lua
+++ b/helpers.lua
@@ -299,6 +299,46 @@ function log_access(user, uri)
   end
 end
 
+-- Check if the request contains the admin token
+function is_admin()
+
+  local admin_token_header = ngx.req.get_headers()["SSOwat-Admin-Token"]
+
+  if admin_token_header == nil then
+      return false
+  end
+
+  -- FIXME : ideally we should also be checking that the permissions for this file are something like 600 + appropriate owner...
+  local admin_token_file = io.open(admin_token_path, "r")
+  if admin_token_file == nil then
+      -- N.B. these messages should use the same syntax as in authenticate()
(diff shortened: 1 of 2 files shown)


helpers.lua
```diff
@@ -299,6 +299,46 @@ function log_access(user, uri)
...
+
+  -- FIXME : ideally we should also be checking that the permissions for this file are something like 600 + appropriate owner...
+  local admin_token_file = io.open(admin_token_path, "r")
+  if admin_token_file == nil then
+      -- N.B. these messages should use the same syntax as in authenticate()
+      ngx.log(ngx.ERR, "Connection failed for: admin")
+      logger.error("Authentication failure for user admin from "..ngx.var.remote_addr)
+      return false
+  end
+  local token = admin_token_file:read("*all")
```
zamentur, line 319: What about timing attack here ?

alexAubin: What do you mean, how do you time-attack a file reading ?

helpers.lua
```diff
@@ -299,6 +299,46 @@ function log_access(user, uri)
...
+      ngx.log(ngx.ERR, "Connection failed for: admin")
+      logger.error("Authentication failure for user admin from "..ngx.var.remote_addr)
+      return false
+  end
+  io.close(admin_token_file)
+
+  local token = token:gsub("\n","")
+
+  -- FIXME FIXME FIXME - this is a really stupid and time-attack-prone way to validate the token
+  if admin_token_header == token then
```
zamentur, line 333: So i think this time attack should be fixed before to merge this. A way to fix that could be to create hash from the token given in request before to compare it

alexAubin: Plot twist, apparently `str1 == str2` is constant time because of the way lua works ...

https://poprocks.dev/constant-time-string-comparison-in-lua/
https://news.ycombinator.com/item?id=6349758

PR #189, opened 2021-06-29
[fix] dash filename, mime types, ynh_userinfo.json

zamentur:
## The problem
I am creating an advanced themes with nice top menu, etc. I discovered several issues:

- Some mime types are missing + we can't specify mime type of .ms
- filename with a dash are not served (and the regex that create this situation might be a security weakness)
- ynh_userinfo.json is broken
- all json are replaced by get_data_for() content ...
- a security risk due to a mechanism that add each uri reached if the app is protected. It was just to display one time an info log (that is already trigger in debug mode...).
- https://github.com/YunoHost/issues/issues/1830

## The solution
This PR fix all of these issues

This PR is found by CNAM

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -208,6 +208,7 @@ end
 if is_logged_in then
     assets = {
                    ["/ynh_portal.js"] = "js/ynh_portal.js",
+                   ["/ynh_userinfo.js"] = "ynh_userinfo.json",
                    ["/ynh_overlay.css"] = "css/ynh_overlay.css"
              }
     theme_dir = "/usr/share/ssowat/portal/assets/themes/"..conf.theme
@@ -218,7 +219,7 @@ if is_logged_in then
     pfile:close()
 
     for shortcut, full in pairs(assets) do
-        if string.match(ngx.var.uri, "^"..shortcut.."$") then
+        if ngx.var.uri == shortcut then
             logger.debug("Serving static asset "..full)
             return hlp.serve("/yunohost/sso/assets/"..full, "static_asset")
         end
--- a/access.lua
+++ b/access.lua
@@ -208,7 +208,7 @@ end
 if is_logged_in then
     assets = {
                    ["/ynh_portal.js"] = "js/ynh_portal.js",
-                   ["/ynh_userinfo.js"] = "ynh_userinfo.json",
+                   ["/ynh_userinfo.json"] = "ynh_userinfo.json",
                    ["/ynh_overlay.css"] = "css/ynh_overlay.css"
              }
     theme_dir = "/usr/share/ssowat/portal/assets/themes/"..conf.theme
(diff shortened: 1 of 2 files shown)


access.lua
```diff
@@ -208,17 +208,18 @@ end
 if is_logged_in then
     assets = {
                    ["/ynh_portal.js"] = "js/ynh_portal.js",
+                   ["/ynh_userinfo.json"] = "ynh_userinfo.json",
                    ["/ynh_overlay.css"] = "css/ynh_overlay.css"
              }
     theme_dir = "/usr/share/ssowat/portal/assets/themes/"..conf.theme
-    local pfile = io.popen('find "'..theme_dir..'" -type f -exec realpath --relative-to "'..theme_dir..'" {} \\;')
+    local pfile = io.popen('find "'..theme_dir..'" -not -path "*/\\.*" -type f -exec realpath --relative-to "'..theme_dir..'" {} \\;')
```
kay0u, line 215: Can you explain this line please?

access.lua
```diff
@@ -208,17 +208,18 @@ end
 if is_logged_in then
     assets = {
                    ["/ynh_portal.js"] = "js/ynh_portal.js",
+                   ["/ynh_userinfo.json"] = "ynh_userinfo.json",
```
kay0u, line 211: What's the purpose of this file and this line?

helpers.lua
```diff
@@ -570,9 +573,10 @@ function serve(uri, cache)
     elseif ext == "ms" then
         local data = get_data_for(file)
         content = lustache:render(content, data)
-    elseif ext == "json" then
+    elseif uri == "/ynh_userinfo.json" then
```
kay0u, line 576: Why don't we do the same thing for other json files?

kay0u (commented):
The code looks good to me, nothing to report so far except that I still can't reproduce the Yunohost/issues#1830

access.lua
zamentur: /ynh_userinfo.json is a file listed in several place of ssowat code (before this PR) (sometimes it's also called ynhpanel.json). Here i just make things working.

The goal is to return a json with data you could use to build a view. For example it contains the list of apps tiles the user can see. In my use case (reviving menu_ynh concept) i could use this to filter/disable menu item.

helpers.lua
zamentur: Cause if i want to put a .json in a theme containing other kind of data i don't want to get the context data but my original json file.

access.lua
zamentur: It's just to avoid listing and serving .git files in the case we make a `git clone` of the theme.

zamentur:
About https://github.com/YunoHost/issues/issues/1830 i reproduce it on 3 servers:

- with my personal server (HDD) 
- with a production VM with SSD. I didn't try with other kind of setup
- with sans-nuage.fr

alexAubin (approved):

kay0u (approved):

PR #190, opened 2021-07-29
[fix] Not enough random file name

zamentur:
Fix 16702 (on forum)

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -713,8 +713,7 @@ end
 -- Read result of a command after given it securely the password
 function secure_cmd_password(cmd, password, start)
     -- Check password validity
-    math.randomseed( os.time() )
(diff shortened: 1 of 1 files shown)


Psycojoker:
LGTM, I think we can merge it if it has been tested.

helpers.lua
```diff
@@ -713,8 +713,7 @@ end
 -- Read result of a command after given it securely the password
 function secure_cmd_password(cmd, password, start)
     -- Check password validity
-    math.randomseed( os.time() )
-    local tmp_file = "/tmp/ssowat_"..math.random()
+    local tmp_file = os.tmpname()
```
alexAubin, line 716: For anyone wondering, the permissions for the file created appears to be 600, which is safe :+1:

PR #191, opened 2021-07-29
[fix] Avoid redirection on unmanaged domains

zamentur:
Fix 16703 (on forum)


helpers.lua
```diff
@@ -1058,6 +1050,15 @@ end
...
+    -- For security reason we don't allow to redirect onto unknown domain
+    -- And if `uri_args.r` contains line break, someone is probably trying to
+    -- pass some additional headers
+    local domain = url:match("^https?://([%w%.]*)/?")
+    if string.match(url, "(.*)\n") or not is_in_table(conf["domains"], domain) then
+        logger.debug("Unauthorized redirection to "..url)
+        flash("fail", t("redirection_error_invalid_url"))
+        url = conf.portal_url
+    end
     return ngx.redirect(url)
```
alexAubin: It looks like in some cases, our `redirect()` is used to redirect to an uri, not an url. such has here https://github.com/YunoHost/SSOwat/blob/dev/access.lua#L57 and 1 or 2 other places ?

alexAubin (changes requested):

helpers.lua
andretheolauret: @zamentur is this PR still under development ?

alexAubin: bumpitybump

zamentur: definitely i will fix that today...

zamentur: should be fixed

helpers.lua
```diff
@@ -1065,6 +1057,15 @@ end
 -- Set cookie and redirect (needed to properly set cookie)
 function redirect(url)
     logger.debug("Redirecting to "..url)
+    -- For security reason we don't allow to redirect onto unknown domain
+    -- And if `uri_args.r` contains line break, someone is probably trying to
+    -- pass some additional headers
```
alexAubin, line 1062: Not tested but I'm also proposing to make sure that the url is prefixed with `https?://`, otherwise the regex may fail to extract the domain

```suggestion
    -- pass some additional headers

    -- This should cover the following cases:
    -- https://malicious.domain.tld/foo/bar
    -- http://malicious.domain.tld/foo/bar
    -- https://malicious.domain.tld:1234/foo
    -- malicious.domain.tld/foo/bar
    -- (/foo/bar, in which case no need to make sure it's prefixed with https://)
    if not string.starts(url, "/") and not string.starts(url, "http://") and not string.starts(url, "https://") then
        url = "https://"..url
    end
```

zamentur: but this doesn't work with uri...

alexAubin: Indeed, I added a `not string.starts(url, "/")`

PR #193, opened 2021-08-31
[Fix] ynh_portal.css: fix font urls

cyrilRomain:


andretheolauret:
@alexAubin there are no errors on my side the path looks ok, 

Ready to be merged I think

alexAubin:
Hmokay but it's not super clear to me how one can reproduce the initial issue ... @cyrilRomain : can you provide context as to how to reproduce the issue that's being fixed here ? It sounds like on all my browsers, fonts are loaded from the woff or raw-data-stuff and not the eot

andretheolauret:
@alexAubin i dont understand like you, I think it’s for when you are in a specific case where you want to proxy or redirect using redirect_ynh the portal 

Exemple maindomain.com/portal/ proxy of maindomain.com/yunohost/sso 

it will not work because the font will return a 404 cause of hard coded path , with this fix it will search font using relative path and will work

cyrilRomain:
@alexAubin you can reproduce the issue by using SSOwat (+ nginx) without installing YunoHost, and by making SSOwat accessible from my.domain.tld/sso. (For learning purpose and because I use some container based apps, I'm trying to setup nginx + SSOwat + LDAP without using YunoHost and see how far I can go).
 
It is worth noting that although SSOwat was done by and for YunoHost, SSOwat is a quite generic suckless SSO and is almost decoupled from YunoHost  (except few links and password.py for password strength checking, which should ideally be done by an LDAP module).

alexAubin:
Interesting, still not sure how that relates to the .eot but yolomerging :thinking:

PR #197, opened 2021-11-14
[fix] unauthorized redirect url check not matching non-alphanumeric chars in domain name

kay0u:
I have a domain like: `do-main.tld`, and this regex doesn't match it.

The new regex "everything except `/`" should works

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -1070,7 +1070,7 @@ function redirect(url)
     if not string.starts(url, "/") and not string.starts(url, "http://") and not string.starts(url, "https://") then
         url = "https://"..url
     end
-    local domain = url:match("^https?://([%w%.]*)/?")
+    local domain = url:match("^https?://([^/]+)/?")
     if string.match(url, "(.*)\n") or (domain ~= nil and not is_in_table(conf["domains"], domain)) then
         logger.debug("Unauthorized redirection to "..url)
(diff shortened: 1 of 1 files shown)


helpers.lua
```diff
@@ -1070,7 +1070,7 @@ function redirect(url)
     if not string.starts(url, "/") and not string.starts(url, "http://") and not string.starts(url, "https://") then
         url = "https://"..url
     end
-    local domain = url:match("^https?://([%w%.]*)/?")
+    local domain = url:match("^https?://([^/]+)/?")
```
kay0u:
```suggestion
    local domain = ngx.var.host
```
🤔

kay0u: Nah this can't work.

We could also iterate through the `conf["domains"]` table to try to match a known domain

helpers.lua
```diff
@@ -1070,7 +1070,7 @@ function redirect(url)
...
     if string.match(url, "(.*)\n") or (domain ~= nil and not is_in_table(conf["domains"], domain)) then
```
kay0u:
```suggestion
    local is_known_domain = false
    for _, domain in ipairs(conf["domains"]) do
        is_known_domain = is_known_domain or url:match("^https?://"..domain.."/?") ~= nil
    end
    if string.match(url, "(.*)\n") or not is_known_domain then
```

PR #204, opened 2022-05-21
Fix auth_header parsing when password contains semicolon

ewilly:
If a colon is in the password, user and password are not well detected.
For example if user="me" and password="pass:word" the function give user="me:pass" password="word"

### DIFF
--- a/helpers.lua
+++ b/helpers.lua
@@ -276,7 +276,7 @@ function validate_or_clear_basic_auth_header_provided_by_client()
 
     -- Try to authenticate the user,
     -- or remove the Auth header if not valid
-    _, _, user, password = string.find(ngx.decode_base64(b64_cred), "^(.+):(.+)$")
+    _, _, user, password = string.find(ngx.decode_base64(b64_cred), "^([^:]+):(.+)$")
     user = authenticate(user, password)
     if user then
         logger.debug("User got authenticated through basic auth")


tomhejra:
Found same issue, fixed it in a similar way but my change was only:
```lua
-    _, _, user, password = string.find(ngx.decode_base64(b64_cred), "^(.+):(.+)$")
+    _, _, user, password = string.find(ngx.decode_base64(b64_cred), "^(..-):(.+)$")
```

which keeps every other aspect the same.
`.+` means: greedy, at least one char
`..-` means: one char, then something shortest possible (even nothing)
       which could be translated to: non-greedy, at least one char

I wanted to create commit for that and open pull request but then I saw this and I don't know how to proceed because I am afraid proposed change by @ewilly  is changing behhavior:

Case ":password":
- current version: returns user nil, password nil
- PR version: returns user "", password "password"   (could be a security problem? "" user may have special meaning)
- my version: returns user nil, password nil

Case ":pass:word":
- current version: returns user ":pass", password "word"   (which is also bad since semicolon is not allowed in usernames)
- PR version: returns user "", password "pass:word"
- my version: returns user ":pass", password "word"

Case "onlyuser":
- current version: returns user nil, password nil
- PR version: error - attempt to perform arithmetic on a nil value (global 'colonindex')
- my version: returns user nil, password nil

EDIT:
Forgot to mention that this started to be an issue for me when updated to 11.0.11 from previous stable version. Since that update I was not able to authorize (HTTP error 401) to some nginx apps like roundcube or wordpress only when I was signed in with this user having semicolon in password. Mentioned diff fixed it for me.

There is a huge security reasoning behind this fix because it inevitably fails and gets logged while also nginx's access error log started logging part of my password before last semicolon in /var/log/nginx/my.domain.com-error.log:
```
2023/01/11 19:39:52 [error] 25978#25978: *105 [lua] helpers.lua:385: authenticate(): Connection failed for: thejra:<part of my password>, client: <my ip>, server: <my domain>
```

ewilly:
Great to see someone else working on that issue  :)
Your proposal fit to my needs.
I made some test here : [Online Lua Compiler](http://tpcg.io/_Y8T4RM)
Once agreed i can close or rewrite the PR
Thanks

tomhejra:
Hmm.. I kinda like the Solution3 but it has other flaws although it handles "" user better (if "" user is allowed).

Solution3 would fail to detect users like: 
- `debian-user-with-hyphens`
- `shadow_utils_or_redhat_user$`
- `redhat.user.with.dots`

I think returning gibberish on "" user is better than not allowing valid usernames. To fix mine to allow "" user it would just be about simply removing one dot:

```lua
-    _, _, user, password = string.find(ngx.decode_base64(b64_cred), "^(.+):(.+)$")
+    _, _, user, password = string.find(ngx.decode_base64(b64_cred), "^(.-):(.+)$")
```

But I did not want to change the explicit `.+` that may have been there for some reason. Normally you would see `.*`.

If I'd add ability to force not-match "" user, We could do for example other simple edit:
```lua
-    _, _, user, password = string.find(ngx.decode_base64(b64_cred), "^(.+):(.+)$")
+    _, _, user, password = string.find(ngx.decode_base64(b64_cred), "^([^:].-):(.+)$")
```
= _"Any character except semicolon + any number of character before semicolon is the user."_ not matching means user = nil, which may or may not mean something different than user "" but it forbids you to use any password for nil/"".

tomhejra:
Regarding this PR I'd keep it to save this conversation about it. You will have to do rebase anyway, so I'd recommend to rebase and `commit --amend` -> `git push --force` in your branch to change content of this PR.

ewilly:
Nice thanks for your advices regarding git command and your lua expertise :)
Apparently I successful rebase and update the commit to your last proposal.
Here to test : [Online Lua Compiler]( http://tpcg.io/_KG3RYV )
If ok for you we can ping alexAudin

helpers.lua
```diff
@@ -276,7 +276,7 @@ function validate_or_clear_basic_auth_header_provided_by_client()
 
     -- Try to authenticate the user,
     -- or remove the Auth header if not valid
-    _, _, user, password = string.find(ngx.decode_base64(b64_cred), "^(.+):(.+)$")
+    _, _, user, password = string.find(ngx.decode_base64(b64_cred), "^([^:].-):(.+)$")
```
alexAubin: Ugh I've read the discussion in diagonal but not sure to understand the whole thing about `([^:].-)` ... not sure to understand why this works ... Why not using `([^:]+)` o_O ?

tomhejra: Correct. `([^:]+)` is equivalent and makes sense... I was blinded by that `.-` approach.

tomhejra: It is completely different look at it, since it means "find as many non-semicolon characters as possible, at least one"but it gets the job done in the same way as `[^:].-`.
Keep in mind that one of major questions here is: Is "" user allowed? This approach forbids it.

alexAubin: hmyeah but `+` also means "at least one char" so it won't match empty string ;P

tomhejra: None of these would match empty string.
Some of those could match empty username (ewilly's first one and `(.-)`) but it was not permited before anyway.

Both `([^:]+)` and `([^:].-)` gets the job done and I am ok with these. `([^:]+)` makes a bit more sense. I don't expect "" user to be something useful anyway, I just wanted to point it out so I got somebody thinking about it because honestly I don't know the answer.

ewilly: commit updated with @alexAubin proposal which is working for me too.

alexAubin (approved):
LGTM, not merging yet just because I don't have time to release the patch for all versions right now ..

tomhejra (approved):

PR #208, opened 2022-11-06
[fix] helpers.lua: openssl v3 support for hmac_sha512

cyrilRomain:
This change is backward compatible with older openssl versions


alexAubin:
Looks legit, cheers !

PR #212, opened 2023-03-10
Add a new setting to redirect unknown URLs to either the SSO portal or a 404 page

Salamandar:
My company would like to disable the automatic "unknown url -> SSO portal" redirection. We would prefer a standard 404 web page.

/!\ UNTESTED /!\

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -315,9 +315,15 @@ for permission_name, permission_infos in pairs(conf["permissions"]) do
     end
 end
 
+---
+--- 5. REDIRECT TO 404 PAGE IF UNKNOWN URL -> PORTAL IS DISABLED
+---
+if not conf["err404_to_portal"] and longest_url_match == "" then
+    return ngx.exit(ngx.HTTP_NOT_FOUND)
+end
 
 ---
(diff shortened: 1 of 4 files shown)


access.lua
```diff
@@ -315,9 +315,15 @@ for permission_name, permission_infos in pairs(conf["permissions"]) do
     end
 end
 
+---
+--- 5. REDIRECT TO 404 PAGE IF UNKNOWN URL -> PORTAL IS DISABLED
+---
+if not conf["err404_to_portal"] and longest_url_match == "" then
```
alexAubin, line 321: Ah that's interesting because URL which do match a permission in the SSOwat conf will still redirect to the portal

Dunno if that was the intent but to me that's indeed the behavior we want ;)

Salamandar: Hmmmm.
Yeah, I think that is. Although… That could very well be configurable too. Do you want the public to know you have a mydomain.tld/very_secret_app_with_a_well_known_name_in_the_url ?

For now, an existing app *and* a non-existing one will have the same behaviour. Maybe some people will want to keep that.
There would be 2 settings :
* Redirect unknown urls to 404 -> default should be true
* Redirect known URLs to 404 instead of login page -> default should be false

PR #217, opened 2023-08-15
SSOwat epic refactoring / Portal API

alexAubin:
c.f. the project to separate : 
- the portal API (login, changing the user stuff like first/last name, mail aliases/forward, password)
- the SSO logic (app permissions, injecting the basic auth header)
- the portal front end (html / css / JS)

This PR removes the old API logic, and tweak stuff to interface with the new portal API in https://github.com/YunoHost/yunohost/pull/1387. The new SSO logic is based on a JWT cookie generated by the new yunohost-portal API.

It also removes the old front end which we probably want to rewrite and manage in a new repo ? For now I just left a super basic `index.html` which demonstrate how the new front-end / API could work.


portal/index.html
```diff
@@ -0,0 +1,59 @@
+<html>
+
```
zamentur, line 2: Missing `<head>`, `<title>` and `<body>` ?
Maybe also add utf-8 and no referrer meta tag (not sure for the no referer)

portal/index.html
```diff
@@ -0,0 +1,59 @@
...
+        <input id="password" type="password" placeholder="Enter Password" name="password" required>
+        </p>
+
+        <button id="login">Login</button>
+    </center>
+
+    <center id="userdata" style="color:green;">
+    </center>
+
+
```
zamentur, line 21: I suggest, if it's easy, to find a way to add a "Legal mention link" here. Indeed adding a Legal mention here help to avoid Google Safe browsing flag.

The idea is to display pages that are not totally identical and without Legal mention.

zamentur (commented):
Thanks for this awesome work !

portal/index.html
alexAubin: This was just a yolo page to test the new API, the actual new portal is being worked on in https://github.com/YunoHost/yunohost-portal

portal/index.html
alexAubin: This was just a yolo page to test the new API, the actual new portal is being worked on in https://github.com/YunoHost/yunohost-portal

But indeed, we should also adjust the API and make this configurable somehow

portal/index.html
zamentur: sorry I'm out of line. Thanks for the info and replying.

portal/index.html
alexAubin: (It would help if you can provide an accurate example, ideally a screenshot, of what you're looking for, like, is it just a link where people could change the `href` target, is it about changing the link text, is it a totally customizable footer ?)

zamentur: For this need, only a simple "Legal mention" link (with translation) that send onto a link that we can configure in the main settings ?

I don't know if a custom footer is really needed. Maybe to add GDPR specific links or return to the main site ? It's also possible that doing that decrease the Flag score, but for the moment changing this was enough for me to be remove from Google Safebrowsing blacklist...

Thatoo:
Will it be possible to configure headers to allow single sign-on, SSO/CAS to be embedded in an iframe?
The idea is to be able to [authenticate by SSO/CAS to synapse from riotchat](https://github.com/gary-kim/riotchat/pull/235) (Element for Nextcloud) without the need to open an other window and make the user feel lost.

alexAubin:
No idea what SSO/CAS is, but that sounds out of scope, the point is not to refactor the SSO per-se to support Open ID connect, SAML or whatever but only to separate the concerns (SSO, portal API, portal front-end)

Josue-T:
> Will it be possible to configure headers to allow single sign-on, SSO/CAS to be embedded in an iframe? The idea is to be able to [authenticate by SSO/CAS to synapse from riotchat](https://github.com/gary-kim/riotchat/pull/235) (Element for Nextcloud) without the need to open an other window and make the user feel lost.

Hello, well normally everything is already implemented here: https://github.com/YunoHost-Apps/synapse_ynh/blob/master/sources/cas_server.php

If the authentication still be too long it the element side issue. I remember there are some option in the element config file to make the authentication more "automatic".

Thatoo:
> > Will it be possible to configure headers to allow single sign-on, SSO/CAS to be embedded in an iframe? The idea is to be able to [authenticate by SSO/CAS to synapse from riotchat](https://github.com/gary-kim/riotchat/pull/235) (Element for Nextcloud) without the need to open an other window and make the user feel lost.
> 
> Hello, well normally everything is already implemented here: https://github.com/YunoHost-Apps/synapse_ynh/blob/master/sources/cas_server.php
> 
> If the authentication still be too long it the element side issue. I remember there are some option in the element config file to make the authentication more "automatic".

I see, so it's not related to SSOwat then but to synapse package.
So I'll open an issue there to ask for this cas_server.php to be able to serve cas within an iframe.

Sorry for the mistake.

PR #218, opened 2023-08-23
[enh] User doc

zamentur:
I suggest the documentation link to redirect onto the user doc and not onto the main doc (used by instances'administrators).


alexAubin:
Hmmmokay but currently this doc page only displays "This section is being worked on. " x_x

zamentur:
Indeed but there are useful subsection to configure email clients.

alexAubin:
Closing because this is terrible UX, many people are just going to think this is a broken link, not realize there are some interesting stuff in the sidebar... We should at least fix the page the link point to with a table of content or something, or explicitly point to the email configuration doc somewhere in the portal

PR #220, opened 2023-09-02
portal-api: Optimization by caching & no check on public routes

selfhoster1312:
After running some benchmarks i found the portal-api branch (with fail2ban disabled) was VERY inefficient due to removing the shared cache for cryptographic computations, compared to the dev branch.

I reintroduced some caching here. Also, i made it so that authentication is only checked on non-public routes.

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -21,6 +21,33 @@ local rex = require("rex_pcre")
 local config = require("config")
 local conf = config.get_config()
 
+-- Cache expensive calculations
+local cache = ngx.shared.cache
+
+-- Hash a string using hmac_sha512, return a hexa string
+function cached_jwt_verify(data, method, secret)
+    res = cache:get(data)
+    if res == nil then
+        logger:debug("Result not found in cache, checking login")
+        -- Perform expensive calculation
+        decoded, err = jwt.verify(data, "HS256", cookie_secret)
+        if not decoded then
+            logger:error(err)
+            return nil, nil, err
+        end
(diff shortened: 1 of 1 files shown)


access.lua
```diff
@@ -21,6 +21,33 @@ local rex = require("rex_pcre")
 local config = require("config")
 local conf = config.get_config()
 
+-- Cache expensive calculations
+local cache = ngx.shared.cache
+
+-- Hash a string using hmac_sha512, return a hexa string
+function cached_jwt_verify(data, method, secret)
```
alexAubin:
```suggestion
function cached_jwt_verify(data, secret)
```

`method` doesn't seem to be used, so let's just assume we're using HS256 everywhere

access.lua
```diff
@@ -55,30 +82,34 @@ end
...
 
     -- cf. src/authenticators/ldap_ynhuser.py in YunoHost to see how the cookie is actually created
 
     local cookie = ngx.var["cookie_" .. conf["cookie_name"]]
+    if cookie == nil then
+        return false, nil, nil
+    end
 
-    decoded, err = jwt.verify(cookie, "HS256", cookie_secret)
+    user, pwd, err = cached_jwt_verify(cookie, "H256", cookie_secret)
```
alexAubin:
```suggestion
    user, pwd, err = cached_jwt_verify(cookie, cookie_secret)
```

(cf previous comment about changing the signature of `cached_jwt_verify`)

PR #231, opened 2024-10-23
Auth headers

Josue-T:
cf: https://github.com/YunoHost/issues/issues/2445

full description here: https://github.com/YunoHost/yunohost/pull/1981

## PR checklist

- [x] PR finished and ready to be reviewed

### DIFF
--- a/access.lua
+++ b/access.lua
@@ -44,13 +44,13 @@ function cached_jwt_verify(data, secret)
         cache:set(data, cached, 120)
         logger:debug("Result saved in cache")
         local headers = { YNH_USER = decoded["user"], YNH_USER_EMAIL = decoded["email"],
-            YNH_USER_FULLNAME = decoded["user_fullname"]}
+            YNH_USER_FULLNAME = decoded["fullname"]}
         return decoded['id'], decoded['host'], decoded["user"], decoded["pwd"], headers, err
     else
         logger:debug("Result found in cache")
         local decoded = json.decode(res)
         local headers = { YNH_USER = decoded["user"], YNH_USER_EMAIL = decoded["email"],
-            YNH_USER_FULLNAME = decoded["user_fullname"]}
+            YNH_USER_FULLNAME = decoded["fullname"]}
         return decoded['id'], decoded['host'], decoded["user"], decoded["pwd"], headers, nil
     end
 end
--- a/access.lua
+++ b/access.lua
@@ -39,14 +39,19 @@ function cached_jwt_verify(data, secret)
         end
         -- As the http_headers is a dictionary and can have any char (we have also the user fullname)
         -- we use the json serializer which is expected to do it correctly
+        -- note we can have special char in the user fullname
         cached = json.encode(decoded)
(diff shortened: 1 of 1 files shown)


access.lua
```diff
@@ -269,6 +276,13 @@ if permission ~= nil and ngx.req.get_headers()["Authorization"] ~= nil then
             ngx.req.clear_header("Authorization")
         end
     end
+
+    -- Clean all Yunohost Authentication Headers
+    for k, v in pairs(ngx.req.get_headers()) do
+        if string.starts(k, "YNH_") then
+            ngx.req.clear_header(k)
+        end
+    end
```
alexAubin: Isn't this block supposed to happen no matter what and not just if an Authorization header is provided ?

Josue-T: The idea there was juste to ensure that there are no way from the client to do user spoofing for all header `YNH_*`. So we remove all header `YNH_*` except the one which come from ssowat.

alexAubin: i understand what this block does, what i'm pointing out is the fact that it's ran only if there's an Authorization header set, in which case the attacker could still spoof the YNH_USER header by not providining an Authorization header

Josue-T: Yes, I understand now. I think in this case, it make more sense to move it out of the if. So we remove in any case the header and then, set it again if the user if authenticated.

access.lua
```diff
@@ -34,18 +35,23 @@ function cached_jwt_verify(data, secret)
...
-        cached = decoded['id']..":"..decoded['host']..":"..decoded["user"]..":"..decoded["pwd"]
+        -- As the http_headers is a dictionary and can have any char (we have also the user fullname)
+        -- we use the json serializer which is expected to do it correctly
+        -- note we can have special char in the user fullname
+        cached = json.encode(decoded)
         cache:set(data, cached, 120)
         logger:debug("Result saved in cache")
-        return decoded['id'], decoded['host'], decoded["user"], decoded["pwd"], err
+        local headers = { YNH_USER = decoded["user"], YNH_USER_EMAIL = decoded["email"],
+            YNH_USER_FULLNAME = decoded["user_fullname"]}
```
alexAubin:
```suggestion
            YNH_USER_FULLNAME = decoded["fullname"]}
```

access.lua
```diff
@@ -34,18 +35,23 @@ function cached_jwt_verify(data, secret)
...
+        return decoded['id'], decoded['host'], decoded["user"], decoded["pwd"], headers, err
     else
         logger:debug("Result found in cache")
-        session_id, host, user, pwd = res:match("([^:]+):([^:]+):([^:]+):(.*)")
-        return session_id, host, user, pwd, nil
+        local decoded = json.decode(res)
+        local headers = { YNH_USER = decoded["user"], YNH_USER_EMAIL = decoded["email"],
+            YNH_USER_FULLNAME = decoded["user_fullname"]}
```
alexAubin:
```suggestion
            YNH_USER_FULLNAME = decoded["fullname"]}
```

PR #232, opened 2024-10-24
Lint Lua with luacheck

Salamandar:
* to be verified, i don't really know lua
* not sure all that is relevant
* my IDE was giving me more errors than luacheck (deprecated stuff, "need to check nil", etc)

### DIFF
--- a/vendor/luajwtjitsi/luajwtjitsi.lua
+++ b/vendor/luajwtjitsi/luajwtjitsi.lua
@@ -68,22 +68,23 @@ end
 -- @param token The JWT token to be parsed.
 -- @return A JSON header and body represented as a table, and a signature.
 local function parse_token(token)
-	local segments=split_token(token)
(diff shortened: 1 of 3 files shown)


vendor/luajwtjitsi/luajwtjitsi.lua
```diff
@@ -26,82 +26,83 @@ end
...
-	end
-
-	return header, body, sig
+    local segments = split_token(token)
+    if #segments ~= 3 then
+        return nil, nil, nil, "Invalid token"
+    end
+    local err, header, body, sig
+
+    header, err = cjson_safe.decode(basexx.from_url64(segments[1]))
```
alexAubin, line 77: This is gonna be confusing if we ever update this vendor lib because we're not the author of this ...

Salamandar: Ah !

